PR #11, opened 2016-06-02
Switch submodule array in init.sh from a static array

tom--pollard:
The submodule array in init.sh was a hardcoded array, which meant
it would require updating to reflect +- changes to .gitmodules.
This change should lower that maintenance overhead.


chbae:
+1 for me.

PR #12, opened 2016-06-02
local.inc: Support only ipk package format

locust2001:
Remove rpm package format support and only support ipk package format.

Signed-off-by: Changhyeok Bae changhyeok.bae@gmail.com

### DIFF
--- a/gdp-src-build/conf/templates/local.inc
+++ b/gdp-src-build/conf/templates/local.inc
@@ -2,7 +2,7 @@
 
 DISTRO ?= "poky-ivi-systemd"
 USE_GSTREAMER_1_00 ?= "1"
-PACKAGE_CLASSES ?= "package_rpm package_ipk"
+PACKAGE_CLASSES ?= "package_ipk"
 EXTRA_IMAGE_FEATURES ?= "debug-tweaks"
 USER_CLASSES ?= "buildstats image-mklibs image-prelink"
 PATCHRESOLVE ?= "noop"


tom--pollard:
I'm all in favour of supporting rpm over ipk by default. A dev can modify this easily if they require different package management. +1

gunnarx:
I sent an email on genivi-meta-ivi@lists.genivi.org just to see if there is any baseline-related concern to consider here.  I don't expect so but it is good to check.

tom--pollard:
Good call @gunnarx , thanks for that.

jeremiah:
My only issue is that if we use rpm we will certainly pull in GPLv3 licensed software (cpio for example). Pulling in GPLv3 is fine if the final target image is designed to be use as an SDK, but if the final target image is meant to be production vehicle ready then this won't work. 

I think before we decide on what packaging format we use in the production image, we ought determine if it is licensed appropriately.

gunnarx:
Yes, agreeing with license policy (whatever it is at a given time) is an important check to do for all PRs.  It would be worth looking over and defining the process we use so that it is always done.

tom--pollard:
I'd point out that we've always had rpm enabled @jeremiah in GDP

jeremiah:
Thanks @tom--pollard. I think that recent discussions are bringing us to reconsider exactly what the purpose of the GDP is. I think there is building consensus that the output of the GDP should be able to go into a car directly. As such we should remove any components that have GPLv3 (or LGPLv3) dependencies. 

Is it possible for us to create builds *without* enabling rpm?

jeremiah:
@gunnarx I suppose pull requests that bring in new components ought to list any new dependencies and their licenses? Not sure how to enforce this, would likely only be able to be done by convention.

tom--pollard:
We should be able to create builds without the rpm package management enabled. Even without the convenience of RPM, GDP still contains GPLv3, in fact we purposely enforce bitbake to allow it

toscalix:
I oppose to the idea of managing GPLv3 at the commit stage. If we want to produce a new output for a GPLv3 free version of GDP we should be talking about a different output for a different target. Check my proposal about this topic: https://at.projects.genivi.org/wiki/download/attachments/11567339/GDP_viewed_as_a_portfolio_v1_0.pdf?version=1&modificationDate=1465385087188&api=v2

gunnarx:
toscalix: I understand where you are coming from here.  I think you are concerned maintainers may not be able to guarantee this on their own and cannot take the full responsibility.  I think you also would like to introduce automated scans.  For both of these points I agree and I think the maintenance team needs to do what it can realistically do.

Nonetheless, in accepting a commit into *any* open source project I assume we all agree that both the technical content and the license of the code must be a consideration.  Whatever the license may be it's an important consideration for integration, so just like maintainers will keep a look out for the technology they should at least keep a lookout for license of the code, proper DCO/Sign-Off, etc.  That is not the same as taking on the responsibility on their own.

As to your comments about "different output for a different target" - after the next BoD meeting there should be more clarity about direction, application of license policy, as well as possibly different variants of GDP that GENIVI wants to officially support in the future.  I will read your input and highlight that into the discussion.   Let's not have that debate here.

Thanks

gunnarx:
@jeremiah wrote

>I suppose pull requests that bring in new components ought to list any new dependencies and their licenses? Not sure how to enforce this, would likely only be able to be done by convention.

I agree it is not easy.  But I would propose that the dependencies typically will come along in the same commit, either being added or "enabled" in one way or another.  Otherwise the change request would presumably not build in Yocto (or in certain corner cases build, but not run correctly).  So that takes care of listing them.  Then it boils down to checking the entire system as it would look when the commit has been applied.

I think the current license information we primarily have is the one given by the bb recipes and the checking that Yocto can do on that.  As discussed in various fora before, working to improve that information and augment it by source code scanning is worthwhile.

CI is testing the technical buildability of code as we do now, but CI could potentially also run a license check.

gunnarx:
Back on-topic:   

I think removing the alternative to RPM does not make sense at this time, despite this comment:

@tom--pollard wrote:
> I'd point out that we've always had rpm enabled @jeremiah in GDP 

Yes and since this PR is not about introducing RPM but about removing IPK.  So I now voted it down based on the given license information on RPM.  If there is a likely decision coming to remove remaining GPLv3 licensed software then we would need to reinstating the alternative (*some* alternative - I do not know the options).  Until we know this it seems like a risk of wasted effort and having to reverse direction.

tom--pollard:
Ok, this can be left open for the time being

chbae:
To support both RPM and IPK is very wasting developer's server resources (build time and storage). So GDP need to decide one package format.

slawr:
As an aside the linaro toolchain layer meta-linaro-toolchain has a known build problem with gcc 5.2 (YP 2.0) when rpm packaging is used. See [bugzilla #2431](https://bugs.linaro.org/show_bug.cgi?id=2431). They seem reluctant to fix it for some reason, although workarounds are straight forward. For our example rootfs we use ipk.

rzr:
I am just curious why do you prefer .rpm over .ipk ? my opinion is opposite I think yocto distros should prefer ipk because using rpm is restricted compared to a real RPM distro (ie: pre/post/scripts are ran on target)

gunnarx:
Discussed at GDP conference today. Main consensus is we should make a choice.  Several votes for IPK but no 100% consensus.

chbae:
Will discuss this GDP call and will merge soon.

zeenix:
@chbae did we discuss this in the end? I don't recall but i might have forgotten a lot of things during xmas. :)

chbae:
@zeenix  No. we didn't yet.

@gunnarx  Could you review this?

gunnarx:
This is a simple change and if some power users want to change it locally they just adjust their local config.  Of course the future may change but it seems good to go ahead and take a decision here.
+1

PR #14, opened 2016-06-07
Upgrade GDP to yocto 2.0

chbae:
# Task description
meta-ivi 10.0.0 is using yocto jethro.

# Poky
URI:	git://git.yoctoproject.org/poky
branch:	jethro
revision:	fc45deac89ef63ca1c44e763c38ced7dfd72cbe1

# meta-oe
URI:	git://git.openembedded.org/meta-openembedded
branch:	jethro
revision:	ad6133a2e95f4b83b6b3ea413598e2cd5fb3fd90

# Other resources
See meta-ivi 10 wiki page: https://at.projects.genivi.org/wiki/pages/viewpage.action?pageId=11567697
Check the updates that come with the new poky version Jethro

[GDP-212] Upgrade GDP to yocto 2.0 (jethro)


tom--pollard:
Is the proposal to move the core layers (poky, ivi, OE) to Jethro, and then move the remaining layers to Jethro at a later date, i.e meta-qt5 etc?

tom--pollard:
Also, we need to decide how to handle the upgrade process branch wise ( at least until we catch up with meta-ivi), if we were to merge all this to master, targets would break.

chbae:
@tom--pollard  I also updated meta-ivi to v10.0.0 but it should be very carefully as you said.
Could you run CI about this change for all target devices?

tom--pollard:
@chbae to test against other targets would require testing against the respective jethro/2.0 bsp layer branches, so automating this is go.cd will be tricky at least for now

chbae:
@tom--pollard 
Thanks. I'll test on minnowboard and qemux86-64 machine tomorrow.

chbae:
Successfully Build and tested on qemux86-64, minnowboard, and raspberrypi2.
However I got the build error in porter.

tom--pollard:
@chbae tested this PR along with the PR to meta-genivi-dev for qemu. Image built and booted as expected:

'Yocto GENIVI Baseline (Poky/meta-ivi) 10.0.0 qemux86-64 ttyS0'

Not seeing any new behaviour issues over what was already known

gunnarx:
+1.  Just as suggested GDP should be based on the baseline.  Consider it an upstream project and modifications there if necessary.   (If modifications are suggested, keep Baseline Integration Team BIT in the loop so that the team can check that the baseline remains compliant).  

Choice of Yocto/OE branch need to be set primarily by the baseline.

This is exactly what you are doing here, so I can stop talking now.  :)

tom--pollard:
https://at.projects.genivi.org/jira/browse/GDP-212 Has some continuing debate on this PR in general, the biggest point of which is tagging/branching/handling non supports targets

tom--pollard:
I'm currently testing the 'GDP10' changes against a yocto2.0/ivi10 based branch in meta-renesas, reporting in GDP-215

gunnarx:
I think the PR-check is doing its job correctly here.

The build seems to fail on all targets because GDP10 has persistence-client-library version 1.1.0 but there is an append that wants version 1.0.0 inside of meta-genivi-dev.

So it's one of those cases where both modules need update at the same time.
@chbae, Is it possible you built against another version of meta-genivi-dev?

(Click on details link above to see the build, and then on the job "init_and_bitbake" for more info.)
or here is the [direct link](https://go.genivi.org/go/tab/build/detail/GDP-Yocto-QEMU_x86-64-GITHUB-PR/8/build/1/init_and_bitbake)

PR #15, opened 2016-06-07
Add Renesas R-Car E2 Silk Support, including latest meta-renesas

tom--pollard:
[GDP-121] GDP 9 port to Renesas Silk, adapted from Stephen Lawrence


slawr:
Thank you for putting together the pull request from my original patches Tom. I considered doing the same but I just thought it would delay things further.

I have not tried building it yet, but based on a visual review I can not see any problem with the commit. Local.conf and meta-renesas SHA both seem to be correct.

tom--pollard:
As mentioned before we should look to consolidate config/templates by $bsp to reduce repeated content, but this can be done at a later date. It would be cleaner if possible to group by renesas/raspberrypi/intel

tom--pollard:
@slawr Not sure how it slipped through my build test, but the meta-renesas branch switch away from 'experimental' introduced a revert to reintroduce bbappend to the gdp image, however still named genivi-demo-platform.  Any chance you could amend the name? Thanks

slawr:
@tom--pollard Ah yes good catch Tom. The perils of a long period between patch submission and merge ;) Yes I'll change the name once I'm in the office this afternoon. Thanks

gdp-src-build/conf/templates/silk.bblayers.conf
```diff
@@ -0,0 +1,7 @@
+include templates/bblayers.inc
+
+# porter specific layer configuration
+BBLAYERS += " \
+  ${TOPDIR}/../meta-renesas \
+  ${TOPDIR}/../meta-renesas/meta-rcar-gen2 \
```
chbae: I thought that you also need to add "${TOPDIR}/../meta-openembedded/meta-multimedia" for silk and porter.

if not, you got the below error.
ERROR: Required build target 'genivi-dev-platform' has no buildable providers.
Missing or unbuildable dependency chain was: ['genivi-dev-platform', 'gstreamer1.0-plugins-bad-asfmux', 'faac']

tom--pollard: I presume this is stemming from the same gdp image bb append, from the include: https://github.com/slawr/meta-renesas/blob/genivi-9-bsp-1.10.0-weston-1.9.0/meta-rcar-gen2/recipes-graphics/images/genivi-demo-platform.bbappend

As I mentioned before my build didn't have this issue, although I think I know why. I may have checked out that sha of meta-renesas, and then re-ran init.sh at somepoint, which will have set it back to the previous commit

chbae: To get below error, firstly I changed "genivi-demo-platform" to "genivi-dev-platform.bbappend" as @tom--pollard said. And then start build.

ERROR: Required build target 'genivi-dev-platform' has no buildable providers.
Missing or unbuildable dependency chain was: ['genivi-dev-platform', 'gstreamer1.0-plugins-bad-asfmux', 'faac']

tom--pollard: Yep it needs the multimedia path, (at least that gets rid of that error). Again as the append wasn't in the previous commits source, this was missed

slawr:
@tom--pollard Renamed the genivi-demo-platform.bbappend to genivi-dev-platform.bbappend. As it's a straight forward change and not to delay further by testing as well I've gone ahead and pushed it to git rep. You can take the tip of the genivi-9-bsp-1.10.0-weston-1.9.0 branch.

gdp-src-build/conf/templates/silk.bblayers.conf
slawr: This appears to be a change introduced by the GDP maintainers in their structural changes. The OE multimedia layer was present in the old structure (see porter branch http://git.projects.genivi.org/?p=genivi-demo-platform.git;a=blob;f=gdp-src-build/conf/bblayers.conf;h=dc8c0d4930044434b8ef6d9906b15154692704ce;hb=refs/heads/porter).

I can see how it might have occurred. During the early days of GDP-ivi9 that used my experimental ivi9 branches I temporarily removed the genivi-demo-platform.bbappend that pulled in the multimedia packages as GDP did not make use of multimedia acceleration and at that point I had not fully tested it. I guess one of the maintainers used what was needed at the time rather than the confs from the porter branch and hence it was moving. Re-enabling multimedia  was one of the improvements I was seeking to introduce with the Silk/Koelsch update I pushed to the ML before the Paris AMM.There it was ok because the confs were correct.

tom--pollard: @slawr I thought that might be the issue, but I can successfully build porter from the single branch without it complaining about missing the multimedia else I would have caught this sooner. Your explanation makes sense to me.

tom--pollard:
Updated the PR to fix the known issues

tom--pollard:
There is still the known issue with Silk and AM PoC tracked here https://at.projects.genivi.org/jira/browse/GDP-170
However I do not think this should stop the merge

PR #16, opened 2016-06-08
bblayers.inc: Add developer.conf

chbae:
When developers run "source init.sh [device]", local.conf.[device] file
is copied from templates' to local.conf. So Before running the script,
developers change the local.conf file, the changed line in local.conf
is lost. To prevent this situation, this idea is suggested.
### How to use

Developers can create gdp-local.conf in gdp-src-build directory and
add some changes in gdp-local.conf instead of local.conf.
hen changes are applied in GDP build.

[GDP-227] Improve single branch of genivi-dev-platform

Signed-off-by: Changhyeok Bae changhyeok.bae@gmail.com

### DIFF
--- a/gdp-src-build/conf/templates/bblayers.inc
+++ b/gdp-src-build/conf/templates/bblayers.inc
@@ -27,3 +27,8 @@ BBLAYERS_NON_REMOVABLE ?= " \
   ${OEROOT}/meta-yocto \
   ${OEROOT}/../meta-ivi/meta-ivi \
   "
+
+# Include local developer settings. This can be used to locally modify the
+# setup of mirrors, caching and any other bitbake variables.
+# The include statement has no effect if the file does not exist.
+include ${TOPDIR}/developer.conf
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ gdp-src-build/tmp/
 gdp-src-build/logs
 gdp-src-build/tmp-glibc
 gdp-src-build/toaster*
+gdp-src-build/developer.conf


chbae:
In my case, I set DL_DIR variable to one directory in local server and share downloads directory for all build.

tom--pollard:
I don't really have a view point here, personally I think it's reasonable for a dev to add to local.conf once the script has generated it. However I don't have any reason to be against this change, so I'm happy to +1 it

gunnarx:
Interesting idea but I think it needs a little more work. 

There are two main configuration files for Yocto right?  local.conf and bblayers.conf.   Ideally we would like this to be a "local" addition to the standard files but the name "local" is already used.  So you suggest "gdp-local" but this name does not say much.    Maybe something like "developer-local" is a little better.

However what I think will be confusing is that you are including "gdp-local.conf" into "bblayers.inc", not into "local.inc".

So can we clarify?  I will call it _my-own-file_ for now.   
First, what happens if the file is not there?  I assume bitbake will fail?  If that is true we must fix it so that the default works.

1. Can make bitbake include _my-own-file_ only if it exists?

2. Or is the idea that we provide an empty <my-own-file> per default?  (It was not part of the commit as far as I can see).

chbae:
@gunnarx 
That's really good review point. I'll check again about naming and included in local.inc 

Openembedded uses two directives (require, include) to include .inc files.
require: the file should be existed in the location.
include: the file is not required (if the file is not existed, openembedded will be ignored)

Thanks

gunnarx:
OK, if include is ignored when the file is not existing, that's OK.

I would suggest to put a comment above the line to indicate this:

# If you have provided some personal settings, they can be added to this file:
include "<some-name>"

OK, so let's consider the name.  Is the intention to include "bblayer.conf" additions or "local.conf" additions?  

(What happens if you mix this content.  Inside bitbake is bblayer.conf and local.conf merged into one big configuration so it does not care, or do you have to keep them separate.  I have never considered that before...    If I added BBLAYERS += something in my local.conf, will it also work?  But I do not recommend it of course)

chbae:
1. Both bblayer.conf and local.conf are working. If I add in local.conf, I'll add it in end of line.
2. About naming,
- dev-local.conf
- developer-local.conf
- etc

gunnarx:
@chbae do you think we need this idea only to extend local.conf or also to extend bblayers.conf?

chbae:
@gunnarx 

extend local.conf is fine to me. if you and other maintainers agree, I'll amend this PR.
Thanks.

gunnarx:
> extend local.conf is fine to me. if you and other maintainers(@tom--pollard, @RobertAJMarshall) agree, I'll amend this PR.

Sorry I didn't understand exactly what you mean by "fine to me".  

I think the current proposal seems to be for modifying local.conf.  Let me check your intention.  The type of changes that the developers will make are:

- Setting DL_DIR to new value
- Set SSTATE_DIR to new value
- Set up location of some local source mirrors.
- Inherit rm_work
...

Your description:
"Developers can create gdp-local.conf in gdp-src-build directory and add some changes in gdp-local.conf instead of local.conf.  Then changes are applied in GDP build."

It all makes sense.  But if the purpose of this is to add changes to the "local.conf", why is the file included by bblayers.inc?  Logically, it should be included by local.inc, don't you agree?.

chbae:
@gunnarx 
My "local.conf is fine to me" means that "dev-local.conf"'s include point.

Logically, you are right. However, GDP is using local.conf for setting per different targets.
So if developers are chaning their setting in local.conf, it's possible to conflict after "git pull".
Also if we are using "git reset --hard HEAD^^" command, developer settings are removed.

tom--pollard:
I'm happy with this, but I still share @gunnarx 's point that it should be in the local.inc instead of bblayers.inc @chbae ?

gunnarx:
Maybe if we merge #23 first then chbae can modify this PR based on that? 

In the new version I would introduce the include -> require change for the other includes.
Because this file is clearly optional, the others are not.

My vote for name is developer-local.conf

gunnarx:
>However, GDP is using local.conf for setting per different targets.
So if developers are chaning their setting in local.conf, it's possible to conflict after "git pull".

I think we are in agreement and I like the original idea.  But we are misunderstanding each other.  
I can see now how my last message was confusing.   By modifying I really meant "modifying by including another file".  

Anyway let's change the PR so we understand each other clearly.   I am proposing:

1. local.inc:
Add:  "include developer-local.conf"

2. local.conf already does include on local.inc.  But now let us change all templates to use require for the mandatory includes.  For example:

minnowboard.local.conf:
"include local.inc" -> "require local.inc"

Can you modify the PR with a new proposal, chbae?  Then it's easy to see if we all think the same thing.

tom--pollard:
https://github.com/GENIVI/genivi-dev-platform/pull/23 is merged now.

chbae:
This is discussed in http://www.openembedded.org/wiki/OEDEM_2016 and will be discussed in openembedded architecture mailing.

slawr:
@chbae looks interesting. Any minutes from that meeting do you know?

chbae:
@slawr 
Still on ongoing and you can see meeting notes in real time in https://docs.google.com/document/d/1fDu0lw_BRh0fNXBByqM6OD_iiHwlsoA9D185Xpe1cUg/edit?usp=sharing

gunnarx:
Discussed on today's GDP conference.  Consensus around creating and including developer-local.conf, and maybe developer-bblayer.conf (if needed).  @chbae will think about it and make a new proposal.

chbae:
Why don't we add just developer.conf? It contains local.conf and bblayers.conf.

gunnarx:
If that works, then I agree, why not?  Could we include it from local.inc then?

chbae:
bblayers.conf is higher level than local.conf. So I thought that if we add bblayers.conf in developer.conf, it's good to add in bblayers.conf.

chbae:
@gunnarx Could you review this?

zeenix:
@gunnarx the ball is in your court I believe. :)

gdp-src-build/conf/templates/bblayers.inc
```diff
@@ -27,3 +27,5 @@ BBLAYERS_NON_REMOVABLE ?= " \
   ${OEROOT}/meta-yocto \
   ${OEROOT}/../meta-ivi/meta-ivi \
   "
+
```
gunnarx: I would suggest a comment here to explain the usage:

> \# Include local developer settings.  This can be used to locally modify the
> \# setup of mirrors, caching and any other bitbake variables.
> \# If the file does not exists the include statement has no effect.
> include ${TOPDIR}/developer.conf

gunnarx:
Sorry.  Looks good.   I'll add my +1, but you might consider my comment.  
You decide, @zeenix or @chbae

zeenix:
Thanks @gunnarx. I +1 your comment so would like @chbae to fix that first. :)

gdp-src-build/conf/templates/bblayers.inc
chbae: @gunnarx 

That's really good. Thanks. I'll modify it.

PR #17, opened 2016-06-10
Update to meta-genivi-dev submodule

tom--pollard:
[GDP-15] - Initial inclusion of FSA


tom--pollard:
The go.cd pipeline failures are not related and are being investigated

PR #23, opened 2016-07-05
Much reduced repetition in configuration files

gunnarx:
Even with the introduction of the local.inc there is still a lot of similarity in the files for  different hardwares.   This started with the DISKMON settings that seemed to be repeated in all local.conf files.  At a minimum, I felt they should go into local.inc if those are settings we want.

But it turns out Raspberry Pi 2 and 3 are building almost exactly the same configuration also - at least right now.  So I created a common file for those.

Then the same for silk+porter.

A little bit of cleanup of unnecessary parts done also.
(@slawr if you want rm_work, I think we should add it back to local.inc.  I removed for now.
But on the other hand chbae has a proposal for developer-local settings in a special file, so that's where you would put it if that proposal is accepted)

So I'm aggressively applying [DRY principle](https://en.wikipedia.org/wiki/Don't_repeat_yourself) here, plus just a little cleanup.  

If any particular commit is not appreciated, let me know and I'll adjust the PR.

I think @RobertAJMarshall, @tom--pollard, @chbae, and @slawr need to review and approve.


gunnarx:
Oh, forgot to say that it is totally untested :-)
So I will let the PR builds do their magic and adjust if it fails.  (...when it fails. ;-)

chbae:
This PR is really nice to me. I'll review soon.

Thanks @gunnarx

gunnarx:
Don't you ever sleep @chbae ?

chbae:
@gunnarx Now' it's 8:35 AM and I'm at office :)

@slawr
About rm_work, I prefer to removing in development side. rm_work is very helpful to reduce storage in CI.

gunnarx:
I think 2 successful and 2 failed builds seems strange.  The failure is a bitbake "sanity check" regarding CONF VERSION?   I need to see if it is a local problem or sstate/cache needs to be wiped to avoid this bitbake complaint.

slawr:
Re rm_work you will notice it is commented out by default. It's just a hint to save people space but if you want to move it elsewhere that is ok by me.

Re commit 4515914 I would rename renesas.local.inc to something like renesas-rcar-gen2.local.inc as it will be different for Gen 3.

Otherwise looks ok based on visual review.

gunnarx:
Thanks slawr.  I noticed it was commented.  For now I'm proposing removal but if we want it included, I would make it more general by putting it in local.inc, but I would rather wait for developer-local proposal from chbae.

4515914 - sure I can rename, or the rename could happen when Gen3 is introduced.

gunnarx:
Something is fishy with PORTER and RPI2 checks.  The only difference I can see is they are the ones that have two levels of includes:
For example local.conf -> raspberrypi.inc -> local.inc
and the two others have only one level.  So if that triggers some bitbake bug maybe...

Still needs investigation if it is this is a confirmed failure or if there is some sstate-cache issue on the Go builds.  Will try a non-CI build to confirm.  

Please hold off on merging until all checks are OK.

gunnarx:
Fixed bugs (I hope)  in new commits a250f69 and ff5bbad.  The other two commits are the same as before rebased on top of the modified commits for R-Pi and Renesas.

The bug was indeed related to the two-level include but not the way I thought.

The root cause was local.inc was not included at all.   It turns out that for local.conf -> include templates/renesas.inc, further includes must be relative to renesas.inc, not to the original file.
So it should be include "local.inc", not include "templates/local.inc"

The reason sanity check complained on the build was that CONF_VERSION was not set at all when local.inc was not included (that has been moved to local.inc since it is always the same)

Bitbake Sanity check could be a bit better and report that CONF_VERSION in fact had *None* value, instead of the standard message "your file was created from a newer/older version".


Some things I learned:

I mimicked the original files here which use **include** but I think we should change it to **require**.  Chbae told me another time that **include** is silent if the file does not exist, but **require** will properly fail if the file is not there.  Had I used require for local.inc, the bug would have been found sooner.

@tom--pollard Let's see that the tests go well.  After that do you want to merge as-is and change to require later, or should I fix that too.

gunnarx:
All checks were OK after bugfix.

gunnarx:
Rebased the series on latest master and force-updated the branch:
 79fe27d -> 2bd79db 
 a250f69 -> 66b30cb 
 ff5bbad ->  5040c24 
 36edb73 -> cfa58fa 
 383d610 -> a33fc07 

(Only meta-ivi URL to Github + faytech fix in m-g-d was new, so I expect no impact)

tom--pollard:
Thanks for this consolidation effort @gunnarx . The relative paths had tripped me up initially when I did the initial conf file structure (mv'ing up a dir during init.sh). I think require could be useful here instead of include but I don't think it's essential. I do think at some point a slight overview of how we handle configuration should be added to README, if it's non self explanatory. +1

PR #24, opened 2016-07-11
Update to meta-ivi 10.0.1

tom--pollard:
meta-ivi
43ac797: Bump version to 10.0.1


chbae:
I thought that you need to change audiomanager_7.0.bbappend in meta-genivi-dev.
In there, you may change SRC_URI_remove.

tom--pollard:
I will be amending this PR with some further dialogue soon

tom--pollard:
Ok, PR created to fix the audiomanager build error https://github.com/GENIVI/meta-genivi-dev/pull/19 (thanks @chbae for the pointer here) which has to be merged before this. Once merged I will add updating the meta-genivi-dev sha to this PR.

One thing of note here, when building I received a QA issue from audiomanager:
>
ERROR: QA Issue: audiomanager: Files/directories were installed but not shipped in any package:
  /usr/lib/cmake
  /usr/lib/cmake/audiomanagerConfig.cmake
Please set FILES such that these items are packaged. Alternatively if they are unneeded, avoid installing them or delete them within do_install.
audiomanager: 2 installed and not shipped files. [installed-vs-shipped]
>

I believe this may stem from https://github.com/GENIVI/meta-ivi/commit/8421336085769112d7abbf228fadffb6e0f018b9 which is included in ivi10.0.1, as we're still building the same commit of AM through the use of bbappend

tom--pollard:
@chbae I'm wondering if you have seen this QA issue when moving to AM7.4 in meta-ivi11?

tom--pollard:
Tested this on both porter & qemux86-64, same QA error with Audiomanager. Even though we override the src uri and ref in m-g-d.git bbappend we still use the same sha1 as the meta-ivi recipes, as such I assume this QA error is also flagged when building the baseline

tom--pollard:
Ok, I've built the 'vanilla' audimanager from meta-ivi10 from scratch without using meta-genivi-dev. The QA error was not present. The meta-genivi-dev uses the same ref as I've already stated, so currently I believe there is something else applied in the append raising the QA issue. Currently we add the Audio Manager plugins through the append (not in meta-ivi10, but is in ivi11) so this may be the cause.

Now, this doesn't have to stop this PR being merged (small QA issue) as it can be rectified later, this PR will also allow the dragonboard PR to be merged as that requires the multilib 64bit patch.

tom--pollard:
Tracked down the offending config:
'
FILES_${PN}-dev_remove = "${libdir}/*"
FILES_${PN}-dev += "${libdir}/${PN}/cmake"
'

The _remove leads to the QA issue

These 2 lines were added to the bbappend in commit
https://github.com/GENIVI/meta-genivi-dev/commit/1cfd9a802c2271e7ba9a082bf267e4567d95d617

@chbae What is your opinion here, is it safe to remove with the changes upstream in ivi10.0.1?

tom--pollard:
Updated https://github.com/GENIVI/meta-genivi-dev/pull/19 with QA issue 'fix'

RobertAJMarshall:
Builds and runs ok for the rasp pi2

tom--pollard:
I plan to merge both PR's today, with no objections

PR #26, opened 2016-07-21
Add DragonBoard-410c EULA acceptance to init.sh

tom--pollard:
Require users to pass 'accept-eula' arg when initialising
build environment for dragonboard-410c


RobertAJMarshall:
Works for me!

jeremiah:
I'm not sure about this and would like to understand better. Is the EULA presented before one gets GPL'd software? Because you cannot put any preconditions to GPL source code, this would definitely included EULAs.

PR #31, opened 2016-07-28
Updates to meta-ivi & meta-geniv-dev submodules

tom--pollard:
meta-ivi includes hotfix 6005751 to node-startup-controller src_url
+ merged PR's to weston & audiomanagerplugins which meta-genivi-dev
is updated to align with.


tom--pollard:
Once this is tested on master, I will be doing the same test on the beta release branch to generate to release artifacts. I will however have to recreate this commit separately as the base sha for meta-genivi-dev is different between the two branches (for this I will not send a PR)

tom--pollard:
Now that the upstream node-startup git issue is resolved I've been able to build and deploy this PR. Working as expect before (audioamanger monitor etc), so I propose to merge. I've also prepared the same commits onto the beta branch

PR #32, opened 2016-07-28
Add force option to init script

RobertAJMarshall:
[GDP-319] Allow the user to override the QEMU only option
  for builds from master

### DIFF
--- a/init.sh
+++ b/init.sh
@@ -22,10 +22,11 @@ done
 if test -z "$machine" ; then
    echo "An invalid target name was given of $choice, available targets are:"
    printf '%s\n' "${targets[@]}"
-   unset machine choice
+   unset machine choice eula
+   echo "Usage: source init.sh target [accept-eula] [-f]"
    return
 else
-    if [[ "$machine" != "qemux86-64" ]] ; then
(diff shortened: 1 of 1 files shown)


init.sh
```diff
@@ -22,17 +22,19 @@ done
 if test -z "$machine" ; then
    echo "An invalid target name was given of $choice, available targets are:"
    printf '%s\n' "${targets[@]}"
-   unset machine choice
+   unset machine choice eula
+   echo "Usage: source init.sh target [accept-eula] [-f]"
```
tom--pollard: I would only highlight the accept eula flag and the -f flag with in the relevant if statements, however adding the unset eula here is a good addition if the user was to typo the machine choice

RobertAJMarshall: I'd find it difficult to say what parameters there were and in which order without specifying them all (though I can see a reason for omitting the source init.sh bit)

tom--pollard: Yes I suppose that is valid, as it stands I've always presumed first is machine second is eula. I only raised it because the eula is only relevant to dragonboard and -f is only relevant to non qemu (or going forward 'non-supported' machines). This would be easier to implement with user interactions during execution, but then it make it harder to automate.

toscalix:
Fine with me.

PR #36, opened 2016-08-04
minnowboard: Add minnowboard yocto2.1 support for GDP11

tom--pollard:
[GDP-266] Support Intel Minnowboard MAX in Master
meta-intel: upgrade to krogoth release
meta-genivi-dev: faytech patch for 4.4 kernel
init.sh: rework to add minnowboard as supported


tom--pollard:
Updated init.sh to ensure all variables are always unset before return.

RobertAJMarshall:
Looks good and builds for me!

PR #38, opened 2016-08-16
poky: Update submodule to latest Krogoth (2.1) ref

tom--pollard:
'GDP11' initially used the ref listed in meta-ivi README for poky.
Updating to the current tip of poky krogoth brings in many (~150)
commits: http://git.yoctoproject.org/cgit/cgit.cgi/poky/log/?h=krogoth


tom--pollard:
Note the bitbake version used in the krogoth release does not seem to play nicely with older / non krogoth supporting layer. Current it's common to see taskhash mismatch errors especially when unpacking / using pre krogoth supporting layers. This PR should be tested by multiple users before merged.

slawr:
It's up to you guys whether you update now, but remember that the M-0.2 release in a few weeks will likely update again to a later Krogoth SHA. Assuming of course Tolkien does that. Unless you can efficiently get through the update before then you might find yourself updating again almost straight away.

RobertAJMarshall:
Built successfully on my build system for both qemux86-64 and rasp pi 2

tom--pollard:
The next step is to discuss with the meta-ivi maintainers about the process going forward in terms of poky dependency

chbae:
If we don't have any problem, I recommend to merge this.

tom--pollard:
I created https://github.com/GENIVI/genivi-dev-platform/pull/39 to follow meta-ivi supported poky/oe refs

PR #34, opened 2016-08-01
Update meta renesas and local.conf

slawr:
Update to the current upstream Renesas R-Car Gen 2 meta-renesas. Also simplify the Porter and Silk board local.conf thanks to changes introduced in meta-renesas. See commit messages for details.

Confirmed to work with the Porter board with the following build config (the GDP SHA before G11):
Build Configuration:
BB_VERSION        = "1.28.0"
BUILD_SYS         = "x86_64-linux"
NATIVELSBSTRING   = "Ubuntu-14.04"
TARGET_SYS        = "arm-poky-linux-gnueabi"
MACHINE           = "porter"
DISTRO            = "poky-ivi-systemd"
DISTRO_VERSION    = "10.0.1"
TUNE_FEATURES     = "arm armv7a vfp neon callconvention-hard cortexa15"
TARGET_FPU        = "vfp-neon"
meta              
meta-yocto        
meta-yocto-bsp    = "HEAD:fc45deac89ef63ca1c44e763c38ced7dfd72cbe1"
meta-ivi          
meta-ivi-bsp      = "HEAD:43ac797bda5dd72ad89771fc95e25296fc8b8b61"
meta-oe           
meta-filesystems  
meta-ruby         = "HEAD:ad6133a2e95f4b83b6b3ea413598e2cd5fb3fd90"
meta-qt5          = "HEAD:ea37a0bc987aa9484937ad68f762b4657c198617"
meta-genivi-dev   = "HEAD:88aaaea7318ba1c4c4dda494af847c868b392760"
meta-rust         = "HEAD:f13ac9d48ae928b761d7be204fa8f877d41e7099"
meta-oic          = "HEAD:69146eaf8bc05c74c377e731b7e16d82854a4659"
meta-erlang       = "HEAD:4d7eacc8e6593934ed5b0c8abc3d3e9dc339d849"
meta-rvi          = "HEAD:de9d548fe35e2cee8688faaae910b4f6f7fea17e"
meta-renesas      
meta-rcar-gen2    = "genivi-10-bsp-1.10.0:98e3453096c4c38fc4819be6888c83d43903383d"
meta-multimedia   = "HEAD:ad6133a2e95f4b83b6b3ea413598e2cd5fb3fd90"

PR #44, opened 2016-09-02
renesas: add renesas r-car gen 2 support for GDP11

slawr:
Upgrade to meta-renesas YP 2.1 release.
Enable porter/silk board support for GDP Master
branch in init.sh again.

### DIFF
--- a/gdp-src-build/conf/templates/porter.local.conf
+++ b/gdp-src-build/conf/templates/porter.local.conf
@@ -1,6 +1,3 @@
 include templates/renesas.local.inc
 
 MACHINE = "porter"
-
-# Setting for u-boot Wayland
-UBOOT_MACHINE = "porter_vin_config"
--- a/gdp-src-build/conf/templates/renesas.local.inc
+++ b/gdp-src-build/conf/templates/renesas.local.inc
@@ -7,9 +7,7 @@ BB_MULTIMEDIA_KERNEL_MODULE = "fdpm-kernel-module|mmngr-kernel-module|\
 BB_MULTIMEDIA_USER_MODULE = "fdpm-user-module|mmngr-user-module|\
                              mmngrbuf-user-module|omx-user-module|\
 							 s3ctl-user-module|vspm-user-module|libmemcpy"
-BB_MULTIMEDIA_TEST_MODULE = "fdpm-tp-user-module|mmngr-tp-user-module|\
-                             mmngrbuf-tp-user-module|s3ctl-tp-user-module|\
-							 vspm-tp-user-module"
+BB_MULTIMEDIA_TEST_MODULE = "fdpm-tp-user-module|mmngr-tp-user-module|mmngrbuf-tp-user-module|s3ctl-tp-user-module|vspm-tp-user-module"
 BB_GST_PLUGINS = "meta-renesas/meta-rcar-gen2/recipes-multimedia/gstreamer"
 MULTIMEDIA_BB = "${BB_MULTIMEDIA_KERNEL_MODULE}|${BB_MULTIMEDIA_USER_MODULE}|\
                  ${BB_MULTIMEDIA_TEST_MODULE}|${BB_GST_PLUGINS}"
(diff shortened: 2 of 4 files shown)


tom--pollard:
Building as we speak :)

tom--pollard:
Built successfully, deploying.

RobertAJMarshall:
Built and deployed here - not hearing any sound though, I will open a ticket

tom--pollard:
Added a comment to the ticket https://at.projects.genivi.org/jira/browse/GDP-365 re audio. Testing a silk build now

RobertAJMarshall:
Silk builds ok for me

tom--pollard:
Tested on both boards, not seeing anything new beyond the already recorded issues. +1 from me. (Maybe the addition of [GDP-213] GDP Master support for Renesas boards in the commit body or the individual silk/porter tickets for to please the jira gods...)

slawr:
Gunnar made a good comment on PR #34. Ideally the local.conf change from there would be pulled as well - see my reply as to why. Although it looks like you have done your builds so perhaps too late for this round.

tom--pollard:
@slawr Feel free to add that commit to this PR :) I have all the build envs available if I need to sanity check the addition...

slawr:
@tom--pollard ok wasn't sure what github allowed in terms of updates. Done. I'll close #34 as superseded by this one

tom--pollard:
@slawr sorry to be pedantic but could you add a signed offline to the first commit? Thanks :)

slawr:
@tom--pollard hmm the github PR compare mechanism doesn't exactly make that detail clear. Which specific comment are your referring to?

slawr:
@tom--pollard if your referring to "renesas: add renesas r-car gen 2 support for GDP11" then it comes via my git rep, rather than a patch file with the other commit (because it was taken from the earlier PR), hence no sign-off. Similar to many of your own commits or am I missing something?

tom--pollard:
@slawr Normally I'd expect that commit to have a signed off line in the message (might be confusing each other here!) If it's a PR with multiple commits then the top level comment can hold the sign-off, as long as the merge commit is preserved.

slawr:
@tom--pollard I think (hope :) ) I understand you. You're saying you would like commit 674d744 (renesas: add renesas r-car gen 2 support for GDP11) to have a signed off line, like the other commit. Is that right?

My point was that the two commits are products of different workflows. 674d744 was formed using the github pr mechanism of branch comparisons. The commit to my fork was not made with -s to add a signed off line (the default in git). The other commit was cherry-picked from the earlier PR #34 for GDP-10 using git format-patch and applied to my fork with git am. In that case as it was a patch process I added the signed off line. Hence why one is signed-off and the other isn't. My final point was that as far as I am aware Genivi was not requiring a sign off on git commit and I can see this seems to be the case for the majority of commits to the GDP git rep. For example Robert's recent commit [8480c](https://github.com/GENIVI/genivi-dev-platform/commit/8480c94f50918fe554b33a62c18c762546663316) or your commit [17e6](https://github.com/GENIVI/genivi-dev-platform/commit/17e644c8e6f06379a6043290709950989d4b3a9b).

I can add a sign off if you really want it but I would need to recreate the commit as adding the sign-off is going to change it. Also I would suggest the need for a sign off in a git commit workflow would need to be made clear somewhere else a lot of people are going to have this rework.

tom--pollard:
@slawr I honestly don't know the in's and out's of what Genivi require (in terms of my commits not having sign-offs, I've raised that before re maintainers requirement to use it and the answer wasn't very helpful...)  I understand what you mean with traditional patch-process vs github fork workflow, maybe it should be discussed in the next GDP call for clarification. Sorry for any confusion on the matter.

slawr:
@tom--pollard No problem at all Tom. It's not easy for you to detect what workflow was used and I think sign off on a patch makes sense. Yes I would say bring it up. It could do with some clarification and normalisation across the alliance.

PR #45, opened 2016-09-19
Update meta-raspberrypi to avoid taskhash mismatch error

RobertAJMarshall:
[GDP-375] the commit reference there appears to solve this
    recurring issue


chbae:
After merging https://github.com/GENIVI/meta-genivi-dev/pull/38, let's merge with meta-genivi-dev layer change.

PR #47, opened 2016-09-21
Add error message if mistakingly sourced from non bash (e.g zsh)

tardyp:
zsh will give cryptic error message

    init.sh:13: unknown file attribute

better make a good error message if developer forgot to switch shell.

### DIFF
--- a/init.sh
+++ b/init.sh
@@ -1,4 +1,11 @@
 #!/bin/bash
+# Note that this script is supposed to be sourced
+if [ ! -n "$BASH" ]
+then
+echo GDP and Yocto in general only works with bash as the interactive shell.
+echo Please retry with bash
+return
+fi
 
 # Parse target board & eula agreement from args
 


RobertAJMarshall:
I'd have an exit (or return) and then fi - the script currently has internal returns  and also should be reindented if there's going to be an extra level.
Also if you run it from sh/dash you still get an error with the interpreter being unable to find the 'fi' and don't get the error message explanation

tardyp:
Indeed, I also tested tcsh and dash

for tcsh, I couldn't find shell subset magic that would work for both as those are very different shell. I've not searched very long. I am not sure if csh is very popular nowadays (unlike zsh)

for dash as it is not an interactive shell I figured it wasn't worth to bother

I fixed the PR with an early return

tom--pollard:
@tardyp Would it be possible for you to add a sign-off to these changes? 

https://www.genivi.org/contribute

tardyp:
changed the commit author and signed-off to my intel.com email address

PR #51, opened 2016-11-07
Add toaster files to .gitignore

RobertAJMarshall:
[GDP-449] toaster files are currently reported as untracked.
This change rectifies that problem.


tom--pollard:
I've not personally used Toaster for a good while, but it's a good tool to have available for use with GDP for those who prefer a GUI based tool.

PR #59, opened 2017-02-15
chromium: Add meta-browser submodule

jaragunde:
Add the meta-browser layer as a submodule, and add it to bblayers.conf together with meta-gnome, which is required because of chromium dependencies.

These changes need to be combined with this PR for meta-genivi-dev: GENIVI/meta-genivi-dev#53

Chromium would not be built by default yet, this is required to add it to the image:
```
IMAGE_INSTALL_append = " chromium-wayland"

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -38,3 +38,6 @@
 [submodule "meta-qcom"]
     path = meta-qcom
     url = git://git.yoctoproject.org/meta-qcom
+[submodule "meta-browser"]
+	path = meta-browser
+	url = git://github.com/OSSystems/meta-browser.git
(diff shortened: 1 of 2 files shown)


zeenix:
@jaragunde thanks! As I said on another PR today, I don't want to merge things that are not part of the image build. Can you please add it default?

jaragunde:
Last patch sent to GENIVI/meta-genivi-dev#53 adds chromium-wayland to genivi-dev-platform image by default.

gunnarx:
FYI, @jaragunde have been trying to get some relevant patches published [with several attempts](http://lists.openembedded.org/pipermail/openembedded-devel/2017-February/111238.html).  What Herve Jourdain refers to as meta-raspberry mailing list in these emails... according to the OE layer index that seems to mean the Yocto list. (Or is there another one?).  Anyhow, nothing yet on the Yocto list it seems.

PR #62, opened 2017-02-23
Add Renesas R-Car Gen 3 board support

slawr:
- Add submodules for the Renesas R-Car Gen 3 (H3/M3) SoC Yocto BSP
- Add initialisation support for the Renesas R-Car M3 Starter Kit low cost board

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -44,3 +44,9 @@
 [submodule "meta-browser"]
 	path = meta-browser
 	url = git://github.com/OSSystems/meta-browser.git
+[submodule "renesas-rcar-gen3"]
+	path = renesas-rcar-gen3
+	url = https://github.com/slawr/renesas-rcar-gen3.git
+[submodule "meta-linaro"]
+	path = meta-linaro
+	url = https://git.linaro.org/openembedded/meta-linaro.git
--- a/gdp-src-build/conf/templates/renesas-rcar-gen3.local.inc
+++ b/gdp-src-build/conf/templates/renesas-rcar-gen3.local.inc
@@ -11,10 +11,6 @@ MACHINE ?= "m3ulcb"
 #SOC_FAMILY = "r8a7795"
 SOC_FAMILY ?= "r8a7796"
 
-# Add systemd configuration
-DISTRO_FEATURES_append = " systemd"
-VIRTUAL-RUNTIME_init_manager = "systemd"
-
 # Linaro GCC
 GCCVERSION = "linaro-5.2"
 
@@ -31,7 +27,6 @@ MACHINE_FEATURES_append = " gsx"
 MULTI_PROVIDER_WHITELIST += "virtual/libgl virtual/egl virtual/libgles1 virtual/libgles2"
 
 # for Wayland/Weston
-DISTRO_FEATURES_append = " pam"
 PREFERRED_PROVIDER_virtual/libgles1 = ""
 PREFERRED_PROVIDER_virtual/libgles2 = "gles-user-module"
 PREFERRED_PROVIDER_virtual/egl = "gles-user-module"
--- a/gdp-src-build/conf/templates/renesas-rcar-gen3.local.inc
+++ b/gdp-src-build/conf/templates/renesas-rcar-gen3.local.inc
@@ -1,0 +1,143 @@
+# Include generic GDP local configuration
+include local.inc
+
+# Add R-Car Gfx and MMP packages to install
+IMAGE_INSTALL_append_rcar-gen3 = "libdrm-kms kernel-module-gles gles-user-module libgbm-dev packagegroup-gstreamer1.0-plugins weston-bin kernel-module-pvrsrvkm"
+
+MACHINE ?= "m3ulcb"
+
+# This sets the SoC
+# H3: r8a7795, M3: r8a7796
+#SOC_FAMILY = "r8a7795"
+SOC_FAMILY ?= "r8a7796"
+
+# Add systemd configuration
+DISTRO_FEATURES_append = " systemd"
+VIRTUAL-RUNTIME_init_manager = "systemd"
+
+# Linaro GCC
+GCCVERSION = "linaro-5.2"
+
+PACKAGE_CLASSES = "package_ipk"
+
+# add the static lib to SDK toolchain
+SDKIMAGE_FEATURES_append = " staticdev-pkgs"
+
+# Disable optee in meta-linaro layer
+BBMASK = "meta-linaro/meta-optee/recipes-security/optee"
+
+# Enable Gfx Pkgs
+MACHINE_FEATURES_append = " gsx"
+MULTI_PROVIDER_WHITELIST += "virtual/libgl virtual/egl virtual/libgles1 virtual/libgles2"
+
+# for Wayland/Weston
+DISTRO_FEATURES_append = " pam"
+PREFERRED_PROVIDER_virtual/libgles1 = ""
+PREFERRED_PROVIDER_virtual/libgles2 = "gles-user-module"
+PREFERRED_PROVIDER_virtual/egl = "gles-user-module"
+PREFERRED_PROVIDER_virtual/libgl = ""
+PREFERRED_PROVIDER_virtual/mesa = ""
+PREFERRED_PROVIDER_libgbm = "libgbm"
+PREFERRED_PROVIDER_libgbm-dev = "libgbm"
+BBMASK .= "|mesa-gl"
+
+# Enable Multimedia features
+MACHINE_FEATURES_append = " multimedia"
+
+# for gstreamer omx plugins
+LICENSE_FLAGS_WHITELIST = "commercial"
+# for mmp test program
+DISTRO_FEATURES_append = " mm-test"
+
+# for weston v4l2 renderer
+#DISTRO_FEATURES_append = " v4l2-renderer"
+
+# OMX H263 decoder library for Linux (RTM0AC0000XV263D30SL40C)
+#DISTRO_FEATURES_append = " h263dec_lib"
+
+# OMX H264 decoder library for Linux (RTM0AC0000XV264D30SL40C)
+DISTRO_FEATURES_append = " h264dec_lib"
+
+# OMX H264 encoder library for Linux (RTM0AC0000XV264E30SL40C)
+#DISTRO_FEATURES_append = " h264enc_lib"
+
+# OMX H265 decoder library for Linux (RTM0AC0000XV265D30SL40C)
+#DISTRO_FEATURES_append = " h265dec_lib"
+
+# OMX MPEG2 decoder library for Linux (RTM0AC0000XVM2VD30SL40C)
+#DISTRO_FEATURES_append = " mpeg2dec_lib"
+
+# OMX Media Component MPEG4 Decoder Library for Linux (RTM0AC0000XVM4VD30SL40C)
+#DISTRO_FEATURES_append = " mpeg4dec_lib"
+
+# OMX Media Component VC-1 Decoder Library for Linux (RTM0AC0000XVVC1D30SL40C)
+#DISTRO_FEATURES_append = " vc1dec_lib"
+
+# OMX Media Component DivXD Decoder Library for Linux (RTM0AC0000XVDVXD30SL40C)
+#DISTRO_FEATURES_append = " divxdec_lib"
+
+# OMX Media Component RealVideo Decoder Library for Linux (RTM0AC0000XVRLVD30SL40C)
+#DISTRO_FEATURES_append = " rvdec_lib"
+
+# OMX Media Component ALAC Decoder Library for Linux (RTM0AC0000XAALAD30SL40C)
+#DISTRO_FEATURES_append = " alacdec_lib"
+
+# OMX Media Component FLAC Decoder Library for Linux (RTM0AC0000XAFLAD30SL40C)
+#DISTRO_FEATURES_append = " flacdec_lib"
+
+# OMX AAC-LC decoder library for Linux (RTM0AC0000XAAACD30SL40C),
+# AAC-LC 2ch decoder middleware library for Linux (RTM0AC0000ADAACMZ1SL40C)
+DISTRO_FEATURES_append = " aaclcdec_lib"
+DISTRO_FEATURES_append = " aaclcdec_mdw"
+
+# OMX aacPlus V2 decoder library for Linux (RTM0AC0000XAAAPD30SL40C),
+# aacPlus V2 decoder middleware library for Linux (RTM0AC0000ADAAPMZ1SL40C)
+#DISTRO_FEATURES_append = " aacpv2dec_lib"
+#DISTRO_FEATURES_append = " aacpv2dec_mdw"
+
+# OMX MP3 decoder library for Linux (RTM0AC0000XAMP3D30SL40C),
+# MP3 decoder middleware library for Linux (RTM0AC0000ADMP3MZ1SL40C)
+#DISTRO_FEATURES_append = " mp3dec_lib"
+#DISTRO_FEATURES_append = " mp3dec_mdw"
+
+# OMX WMA decoder library for Linux (RTM0AC0000XAWMAD30SL40C),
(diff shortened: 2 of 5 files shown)


slawr:
In creating the PR I see that although I had branched from a recent Master branch commit on 10th Feb there have been a couple of more recent commits including one to add meta-browser. Hence the merge conflict Github reports.

zeenix:
@slawr Thanks. Yeah, you'll need to rebase first. :)

slawr:
@zeenix, @gunnarx what's the official GDP policy in that regard? I'm thinking of avoidance of the case where I now say spend 2 says rebasing and testing, only for you to merge Phil's iot layer additions in the meantime and you then ask me to rebase, and so on and so on.

zeenix:
@slawr That's pretty usual in all open source projects, for contributors to have to rebase their work multiple times before their work is merged. Having said that, if your PR is good, it's unlikely you'll have to rebase more than 2 times and typically there are no conflicts and rebase can happen (automatically) as part of the merge.

slawr:
@zeenix Hi, agreed rebasing is common when looking to push to a distributed project. I was just wondering if the project had any guidelines in place. I know for example what I need to do to improve the chances of a successful merge with our upstream kernel team. So I was just looking to keep cycles down where possible. But if there is no guideline in place that is ok. I'll rebase on the current master head and resubmit tomorrow.

slawr:
@gunnarx you mean Gen 2 vs Gen 3 I think. It's something I am aware of and am discussing internally. Currently they are separate as you have noted and although I could merge them myself I decided not to. It's more important to keep a clean history of how my few commits correspond to the standard Yocto BSP I feel. Regarding Gen 3 we are working to integrate these commits into the next release of the standard Yocto BSP, i.e the standard YBSP will gain Genivi support.

The question of other systems is a wider topic involving many threads and stakeholders, including sync between such systems - as you say we can take that offline.

gunnarx:
Oh, such responsibility.  I hereby decide the policy to be:  **Rebase early and often!**  :-)

Basically, yeah, what he said.  It's ideal to just catch up with master, as many times as you need to until it's your turn to be merged.  If you are seriously uncomfortable with it, maybe a submitter could ask really nicely that the maintainer take this burden on for you during the merge.  It's at the discretion of the maintainer.

But I do understand a concern not only for the rebase step but sometimes also feeling the need to test.  Note that the concern is equal for any contributor.   Like in your example, if you were merged before Phil, he would be in the same position.  So there's no more fair way to do it.

Also, the CIAT system will re-trigger and report its results every time.  So you already get the compilation check, and ... oh, let's have automated tests too, yeah you know the drill.  But unless contributors do the work and update the PR, we won't get those automated checks either.  Delegating to the maintainer to rebase during merging is more risky since I would not expect he wants to/will rebuild locally and test immediately upon each PR merge.

So in summary, to reduce this you just have be *better* than everyone else, so that you get merged first!  Here, supporting a new major hardware should score pretty high I would say, right @zeenix?  Minor bribery of @zeenix with any high-end fancy tea, is also known to work wonders.  (Just kidding/disclaimer/ corporate conduct policy applies...)

zeenix:
@slawr thanks for you understand and thanks @gunnarx for explaining (but not for exposing my weakness in public :)). Regarding maintainer helping out in rebasing, I wouldn't mind at all but last time I tried to fix some commits for a contributor, I didn't find a way to inform github that I've actually merged the PR (with minor modifications). i'm still a bit new to github PRs so maybe I'm missing something here. Can I (maintainers) actually push to the repo the PR is based on? I do see a message "Add more commits by pushing to the gen3-gdp-11 branch on slawr/genivi-dev-platform." Is that really possible? If so, can I force push (I'll need to do that if I do rebase)?

gunnarx:
> I didn't find a way to inform github that I've actually merged the PR

No you are right, that may not be possible, anf after rebase GitHub has no way of figuring it out through the hashes.  I'm probably thinking of a process like patch submissions to a mailing list where you'd just notify that it was accepted.  You could of course just comment on the PR that it has been merged, and close it, but it won't have the GitHub goodness.

zeenix:
Thanks for confirming @gunnarx . That's unfortunate but it's also a good reason for contributors to take care of rebasing. Of course, if they have to rebase too many times, we can suspend this rule and give up on github goodness in such cases. Let's hope such cases are rare.

slawr:
>  I do see a message "Add more commits by pushing to the gen3-gdp-11 branch on slawr/genivi-dev-platform." Is that really possible?

I assume you would need write privileges to my git rep which you don't have. So I assume not.
I did see that github allowed me to fix conflicts in github itself, but that may have been as I raised the PR.

I can't say I'm that expert with githubs PR mechanisms but it seems it has some limitations against the traditional patch process. Adding signed off would be one good example where the github process is less flexible.

@gunnarx ok I see the policy. I agree with your comment about Phil. I don't seek an advantage just any project guidelines that limit rework. I think we can leave it there.

I've rebased my branch on the latest master and forced pushed my branch. It works locally for me on M3 Starter Kit. Interesting that github sees the change but does not archive the history of the first commit in some way.

gdp-src-build/conf/templates/renesas-rcar-gen3.local.inc
```diff
@@ -0,0 +1,143 @@
...
+
+# Disable optee in meta-linaro layer
+BBMASK = "meta-linaro/meta-optee/recipes-security/optee"
+
+# Enable Gfx Pkgs
+MACHINE_FEATURES_append = " gsx"
+MULTI_PROVIDER_WHITELIST += "virtual/libgl virtual/egl virtual/libgles1 virtual/libgles2"
+
+# for Wayland/Weston
+DISTRO_FEATURES_append = " pam"
```
zeenix: @chbae Thanks! Could you please set "requested changes" status on the patch?

gunnarx:
Just for completeness, let me comment that the failing check on Porter is the well-known issue and is unrelated to this PR.  +1 therefore on the build status from me.

gdp-src-build/conf/templates/renesas-rcar-gen3.local.inc
```diff
@@ -0,0 +1,143 @@
...
+MACHINE ?= "m3ulcb"
+
+# This sets the SoC
+# H3: r8a7795, M3: r8a7796
+#SOC_FAMILY = "r8a7795"
+SOC_FAMILY ?= "r8a7796"
+
+# Add systemd configuration
+DISTRO_FEATURES_append = " systemd"
+VIRTUAL-RUNTIME_init_manager = "systemd"
```
slawr: Firstly thanks for the review Changhyeok.

Although it is a duplication I think it would be best to keep it. My reasoning is that local.conf entries like this are pretty portable in describing how you might configure the Yocto BSP to work with a Genivi based distro. So someone might take this a starting point to use in their own platform and that may not be meta-ivi based.

It is also useful to minimise minor differences between platforms. So an R-Car Gen 3 local.conf for GDP may be consistent with one for a different distro.

gdp-src-build/conf/templates/renesas-rcar-gen3.local.inc
```diff
@@ -0,0 +1,143 @@
...
+
+# Linaro GCC
+GCCVERSION = "linaro-5.2"
+
+PACKAGE_CLASSES = "package_ipk"
```
slawr: No the YBSP can support either.

I selected ipk for two reasons. Firstly I thought of the recent request to move Genivi builds to be ipk only. Secondly there is an upstream issue in meta-linaro with gcc 5.2 when rpm is selected that Linaro seem to not want to fix. So selecting rpm requires a meta-linaro patch such as the one attached to workaround the issue:
[0001-libgcc-5.2-Workaround-upstream-build-failure-with-rp.patch.zip](https://github.com/GENIVI/genivi-dev-platform/files/807041/0001-libgcc-5.2-Workaround-upstream-build-failure-with-rp.patch.zip)

Given the first point I chose to solve the second by selecting ipk rather than carry a patch to a different layer.

gdp-src-build/conf/templates/renesas-rcar-gen3.local.inc
slawr: See comment above about systemd

gdp-src-build/conf/templates/renesas-rcar-gen3.local.inc
slawr: @chbae: I can't find a way to reply in line to your top comment about readme and the board setup name so I will add comments here.

Re r-car-m3: My comment assumes you are talking about the init.sh target parameter. I agree that "r-car-m3-starter-kit" is quite long but I found no good shortened form. Abbreviating "starter-kit" to 'sk' for example assumes knowledge about the board. The problem with "r-car-m3" is it only describes the SoC, not the board. We have R-Car M3 Starter Kit and R-Car M3 Salvator-X boards, as well as R-Car H3 SoC variants for example. So I don't think we can move to just r-car-m3.

Re README.md: Yes good catch. I thought about modifying it but I considered we needed to discuss whether this would be a supported board or not first. It could then be updated later, but I can update now if you prefer?

gunnarx:
I would like all lines that are already defined for all hardware to be removed from the Renesas local file.  I would like the Renesas file to describe/modify only what is unique for the Renesas build.

init.sh
```diff
@@ -74,7 +74,7 @@ echo "Local & bblayers conf set for $machine"
 # bsp submodule layer
 declare -A bsparr
 
-bsparr+=( ["minnowboard"]="meta-intel" ["raspberrypi2"]="meta-raspberrypi" ["raspberrypi3"]="meta-raspberrypi" ["porter"]="meta-renesas" ["silk"]="meta-renesas" ["dragonboard-410c"]="meta-qcom" )
+bsparr+=( ["minnowboard"]="meta-intel" ["raspberrypi2"]="meta-raspberrypi" ["raspberrypi3"]="meta-raspberrypi" ["porter"]="meta-renesas" ["silk"]="meta-renesas" ["dragonboard-410c"]="meta-qcom" ["r-car-m3-starter-kit"]="renesas-rcar-gen3")
```
zeenix: Yeah, makes sense but I'd like the change to be part of this PR itself or done before this PR but that's just a preference. The reason is that otherwise we'll either forget and/or lose the motivation, while blocking a PR for it gives motivation to contributor to address the issue right there and then.

.gitmodules
```diff
@@ -41,3 +41,9 @@
 [submodule "meta-browser"]
 	path = meta-browser
 	url = git://github.com/OSSystems/meta-browser.git
+[submodule "renesas-rcar-gen3"]
+	path = renesas-rcar-gen3
+	url = https://github.com/slawr/renesas-rcar-gen3.git
```
zeenix: While I'm very happy and eager to get Gen3 support in GDP, I strongly feel that Renesar layer should handle Gstreamer plugins better, so it does not have to completely override the whole plugin packages and assume a very specific version of gstreamer plugins:

https://github.com/slawr/renesas-rcar-gen3/blob/genivi-10/meta-rcar-gen3/recipes-multimedia/gstreamer/gstreamer1.0-plugins-good_1.4.5.bbappend#L1

If Renesas needs to modify existing gstreamer plugins instead of providing their own, ideally such modifications should be pushed upstream but if that is not possible for any reason, patches against gstreamer plugins in recipes would also be acceptable (at least a lot more cleaner solution than what we have right now).

It would make me very happy as GDP maintainer if this issue is resolved as part of Gen3 addition.

init.sh
gunnarx: I understand your reasoning.  This change doesn't make much sense before the PR though.  The way I planned to create it, it needs modification of the commit that is included here.  So that means it needs to be part of the PR or come later.  I would propoose let's not block this on such a detail.  I'll create a ticket if you think we'll forget :)

.gitmodules
slawr: @zeenix Are you saying you will not accept the PR without your changes? I think the changes would need to be defined in much more detail but that question is the first gate I think.

zeenix: @slawr No but it would be a sad-faced merge. :) I don't think more details should be needed from GDP team. It all depends on how Renesas devs decide to solve the issue.

slawr: @zeenix I will live with your sad-faced merge then.

gdp-src-build/conf/templates/renesas-rcar-gen3.local.inc
chbae: @zeenix I'll do from next time :)

gdp-src-build/conf/templates/renesas-rcar-gen3.local.inc
chbae: Re README.md: Yes good catch. I thought about modifying it but I considered we needed to discuss whether this would be a supported board or not first. It could then be updated later, but I can update now if you prefer?
>> Not exactly now. up to you.

.gitmodules
chbae: I agree with @zeenix but looks fine to me.

gdp-src-build/conf/templates/renesas-rcar-gen3.local.inc
chbae: Thanks for your reply.

slawr:
@gunnarx, @chbae: this comment threading is becoming a bit of a mess :)

As agreed with Gunnar this morning I have removed pam and systemd from the renesas gen 3 local.conf. I did it as a separate commit, rather than squashing it into the existing one, so that readers can see it is a subtraction for GDP only.

I notice that for some reason currently the automated PR build tests are only building for Qemu and minnow board.

gdp-src-build/conf/templates/renesas-rcar-gen3.local.inc
slawr: @chbae Hi I would say then lets update README.md once the PR is settled as the SHA keeps changing as a result of change requests.

init.sh
slawr: @gunnarx good catch Gunnar. I didn't spot that when adding it to the script. 

Just as input to any script redesign meta-linaro provides various linaro developed technologies including the Linaro toolchain. As a result although I've added it is entirely conceivable that other ARM YBSPs that come along later will also make use of it and so could be considered as shared amongst various YBSPs. That said, I suppose it is possible that for some reason they may use different SHAs.

gunnarx: I decided with @zeenix today to add this separately as soon as this is merged.  Setting Review to OK.

gunnarx:
> I notice that for some reason currently the automated PR build tests are only building 
> for Qemu and minnow board.
I have not seen that.   4 targets are building now.  Go.CD will inform GitHub that a check is pending when it starts building (or re-starts due to some change).  Depending on capacity they all might not start immediately so maybe that is what you saw.

init.sh
gunnarx: What you mention is not a problem  - other targets that need meta-linaro can simply also state that they need it.   After I do the script change, it should be clear.

> That said, I suppose it is possible that for some reason they may use different SHAs.

Actually doing that is not easily supported with the single-branch strategy we have now.  But let's deal with the problems we actually do have.

gunnarx:
All checks have rebuilt again.  Porter of course fails because this PR is not based on the Porter fix (i.e. not rebased on latest master as of today).

Time to proceed with sad-faced merge? :)

zeenix:
@gunnarx or time for nitpicking. :) Just a bit.

zeenix:
@slawr Actually you have mostly done a pretty good job with commits so not a lot to nitpick about but just a few simple fixes and we'll have it merged (We don't want the face to be too sad):

* Shortlog/summary
  * needs to be around 50 characters, ideally under 50. Here just "Add Renesas R-Car M3" would be fine and in description you can explain the specifics. 
  * prefix is supposed to be name of a module.

* Description part is kinda independent and there is a mandatory empty line between summary and description.

**Turns out I can push to other's clones. :) So I went ahead and fixed the above issues in commit logs.**

slawr:
Hi, is there anything you are waiting for from me? From zeeshan's last comment in bold I assumed not.

zeenix:
@slawr kinda, i wanted you to check if I didn't break anything but I'll take your last comment as an ACK.

PR #72, opened 2017-03-21
poky: Update to latest krogoth (b4df9df4)

zeenix:
This is to fix https://at.projects.genivi.org/jira/browse/GDP-577 .  The issue unfortunately also affects gdp-11 so we need to apply this on that branch too.

**NOTE**: The fix in question is no longer needed but as said in comments, better update anyway.


chbae:
Looks good for me but I'll wait for CI result and then merge it.

zeenix:
@chbae Yeah, I wouldn't want to merge before that either. :)

chbae:
@zeenix 
Last file commit in upstream is reverted because upstream restored the original hashes.
So we don't need to take this PR.

zeenix:
@chbae Oh good. That's good to hear.

zeenix:
@chbae However, I think we should update to latest stable branch anyway, since there are many fixes. Hopefully we'll soon move to morty, thanks to your awesome work but in case we don't in release 12, we should be on latest krogoth at least.

chbae:
@zeenix Then you need to add 1 more commit (Revert file: update SRCREV for 5.25 to fix fetch fail on missing commit) from krogoth branch of upstream.

zeenix:
@chbae actually I already did but seems I forgot to put that change in commit log. I'll fix that now..

chbae:
Dragonboard failure cause that this commit is not latest but it's fine. I'll merge it.

PR #73, opened 2017-03-23
Add (unofficial) Koelsch board support

gunnarx:
Builds for Renesas Koelsch development board have previously been done
by building for the Porter board instead, and init.sh did not support
koelsch as a choice.  Added separate config files and koelsch as
unofficially supported.


gunnarx:
I added Koelsch to the PR checks for the only reason to test this PR but some other PRs have come in between and get failed checks because the Koelsch parameter is not yet there.  Therefore I'd rather not keep this around any longer than necessary.  I'll go ahead and merge it now - hope it's OK.

PR #79, opened 2017-04-21
Add script to help CI builds

gunnarx:
This works well for me now on my fork.  I'm sure it will need updates but it would help if we can merge it to master quickly and have it available also for the standard CI.

### DIFF
--- a/scripts/ci-build.sh
+++ b/scripts/ci-build.sh
@@ -1,0 +1,299 @@
+#!/bin/bash
+
+# (C) 2017 Gunnar Andersson
+# LICENSE: MPLv2
+
+# User warning!
+
+# A casual user will need to read the script to understand what the first
+# parameter shall be to make it run:
+#
+if [[ "$1" != CI_FLAG ]]; then
+
+  cat <<EOT
+
+READ THIS: This script was not designed for personal use.
+
+If you still use it, you hereby agree to any applicable End-User License
+Agreement (EULA) of the corresponding BSPs that are used.
+
+It is still a useful script but a few things could be destructive if you use
+this in your development environment instead of a "throw-away" environment like
+CI.
+
+Please make sure you read the script to understand what it does!
+
+After that I hope you enjoy using it.
+
+EOT
+
+   exit 2
+fi
+
+# ---- General settings ----
+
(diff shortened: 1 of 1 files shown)


gunnarx:
Updated to add sign-off

gunnarx:
Describing every step would not make sense - it is better read in the script and I thought these lines said enough:

> This helper script takes care of a lot of details to simplify setup of
> GDP builds in CI.  Instead of the steps being defined in the CI tool,
> they are defined here.  

I assume those lines should be understood as the script contains build steps.

Since you want more I decided to describe all the categories of things it does, with examples.  I could not do that any shorter than this.  Perhaps unnecessarily long in my opinion but it should be extremely clear now.  I'll merge it like this because we need it now...

zeenix:
> Describing every step would not make sense - it is better read in the script and I thought these lines said enough:

I never asked you to do that. It's a complicated bash script so a *summary* of what it does would have been nice. If you are going to merge things without acting on PR or waiting on others to agree, there is no point of submitting a PR at all.

> This helper script takes care of a lot of details to simplify setup of
> GDP builds in CI. Instead of the steps being defined in the CI tool,
> they are defined here.

That is extremely vague and hence not extremely useful.

gunnarx:
> I never asked you to do that.

No, I realize that.  But for someone not already understanding the steps involved in the CI-build I could not come up with good specifics in between that vague comment and an overly comprehensive description.

>  If you are going to merge things without acting on PR or waiting on others to agree, there is no point of submitting a PR at all.

Well... that's seeing things black and white.  Couldn't a PR serve as documentation?  Finally, I thought it would give you a better indication that something happened since PRs are notified via emails, and direct pushes are normally not.   It could also be not always possible to know before hand when something needs to be merged when the PR is created.

I'll try to predict when something is urgent and avoid the PR.  I hope you understand this one was urgent.
But I modified based on your first feedback so it served a purpose this time.

zeenix:
> I'll try to predict when something is urgent and avoid the PR. I hope you understand this one was urgent.

I don't know that. From your commit log, it seems like something that will help with CI, not fix something that needs urgent fixing.

PR #84, opened 2017-05-03
renesas-rcar-gen3: Update to 72302d0b to fix GDP-595

slawr:
Update the renesas-rcar-gen3 submodule to fix JIRA ticket GDP-595
(Qtbase do_compile failure: EGL/eglmesaext.h missing).

[1] https://at.projects.genivi.org/jira/browse/GDP-595

Signed-off-by: Stephen Lawrence <stephen.lawrence@renesas.com>


zeenix:
The rcar3 build fails with:

`Cannot read /var/lib/go-agent/pipelines/GDP-Yocto-r-car-m3-starter-kit-GITHUB-PR/gdp-src-build/tmp/work/aarch64-poky-linux/qtbase/5.6.0+gitAUTOINC+cec15a89d0-r0/git/mkspecs/oe-device-extra.pri: No such file or directory`

slawr:
@zeenix: that looks like one of the build order issues @gunnarx mentioned in the email thread. When I look at the build check [above](https://go.genivi.org/go/tab/build/detail/GDP-Yocto-r-car-m3-starter-kit-GITHUB-PR/8/build/1/init_and_bitbake) at the bottom I see the OpenGL do_configure that is [GDP-590](https://at.projects.genivi.org/jira/browse/GDP-590).

So I don't think this update is causing these problems as they have been seen before on the CI. It may indirectly help them as the dependency on Mesa may delay the building of qtbase in the order.

Regarding the issue you list I have not been able to reproduce it locally yet which makes it difficult for me to investigate. Ideally you need to be able to access the build to see what has and has not been built. I would like to discuss that with @gunnarx at the AMM.

zeenix:
ah ok, thanks for explaining @slawr.

PR #85, opened 2017-05-11
init.sh: Make script compatible with zsh

rpannek:
Some people run zsh instead of bash, it's nice to have the script compatible
with both. Thes patch makes this possible and at the same time it makes all
used variables local within a function so they don't leak out without the
need of deleting them at the end.

### DIFF
--- a/init.sh
+++ b/init.sh
@@ -1,103 +1,119 @@
-#!/bin/bash
 # Note that this script is supposed to be sourced
-if [ ! -n "$BASH" ]
-then
-echo GDP and Yocto in general only works with bash as the interactive shell.
-echo Please retry with bash
-return
+
+function setupGitSubmodules() {
+    local choice=$1
+    if [[ $2 == "accept-eula" ]] ; then
+        local eula=true
+    fi
+    echo "You selected target $choice"
+
+    local targets=(
+        "qemux86-64"
+        "koelsch"
+        "porter"
+        "raspberrypi2"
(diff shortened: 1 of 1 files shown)


slawr:
Hi, I don't think the R-Car Gen 3 build failure is connected to this PR. It is one we have been suffering recently on Go.CD. It typically does not show up with local builds but that is immaterial. It's recorded as JIRA [GDP-590](https://at.projects.genivi.org/jira/browse/GDP-590). It is a build order error related to a missing build dependency. I am discussing the correct fix, but a workaround is to build libgbm then restart the build.

gunnarx:
Does it work with any bourne shell?
If not, maybe there should still be a check and a message if you are not running bash or zsh.

gunnarx:
@rpannek - I added the bash or zsh check to warn user in my copy - so I can amend the commit for you.   
But could you first add a sign-off to your patch?  Note also typo: s/Thes/This/

rpannek:
Ok cool, I added the signoff and fixed the typo.

gunnarx:
That's odd - I'm not able now to update the PR by pushing to it (i.e. pushing to rpannek/gmaster)... That always works normally.  Did you add particular branch protection or something @rpannek?

gunnarx:
Never mind I'll just add my change in a new commit.

rpannek:
Not that I know of but cool, thanks for merging!

PR #86, opened 2017-05-18
renesas-rcar-gen3: Update to 4ebc763eb4 to fix GDP-590

slawr:
Update the renesas-rcar-gen3 submodule to fix JIRA ticket GDP-590
(qtbase do_configure build failure: OpenGL ES 2.0 disabled) [1].

To enable the fix add packagegroup-graphics-libegl to IMAGE_INSTALL.

[1] https://at.projects.genivi.org/jira/browse/GDP-590

### DIFF
--- a/gdp-src-build/conf/templates/renesas-rcar-gen3.local.inc
+++ b/gdp-src-build/conf/templates/renesas-rcar-gen3.local.inc
@@ -2,7 +2,7 @@
 include local.inc
 
 # Add R-Car Gfx and MMP packages to install
-IMAGE_INSTALL_append_rcar-gen3 = "libdrm-kms kernel-module-gles gles-user-module libgbm-dev packagegroup-gstreamer1.0-plugins weston-bin kernel-module-pvrsrvkm"
+IMAGE_INSTALL_append_rcar-gen3 = "libdrm-kms kernel-module-gles gles-user-module libgbm-dev packagegroup-gstreamer1.0-plugins weston-bin kernel-module-pvrsrvkm packagegroup-graphics-libegl"
 
 MACHINE ?= "m3ulcb"
 


chbae:
@slawr 

You said "Workaround is to bitbake libgbm, then restart bitbake for genivi-dev-platform" in jira. Then I thought that it's better to add libgbm dependency only for M3 in qtbase.

slawr:
@chbae Hi, thanks for the review. Yes, for qtbase to build libgbm and wayland-kms need to build first and your suggestion would be an approach that should work. It would be simpler, assuming it avoids the circular dependency, but has the downside that it would fix the problem only for GDP. Anyone trying to build something other than GDP could still see the problem.

That said, I see the build check has just failed with the same issue. I have always been unable to reproduce the issue locally building GDP. The only way I can reproduce it locally is by building qtbase instead and that case it fixed. Anyway I will update the JIRA ticket. Looks like this PR will have to be withdrawn.

gunnarx:
@slawr Did you know that you can use [this script](https://github.com/GENIVI/genivi-dev-platform/blob/master/git-submodule-bump) for submodule bumps?  It provides a shortlog of what changed inside the submodule, and you can then amend the commit to add some more details if you feel it's necessary.

Also, instead of "to fix JIRA ticket GDP-590" in the body, remember the defined format of referencing tickets.  Put this line below the body, above the sign-off.

[GDP-590] Qtbase do_configure build failure: OpenGL ES 2.0 disabled

(It's not only visual consistency but also for tool integration.  Yes JIRA seems to pick up references anyway through text-matching, but consistency allows for easy automation).

(Yes the spirit of @zeenix lives on... ;-)

slawr:
@gunnarx ok and thank you for being specific about the points. btw I realised this morning I forgot to sign :)

One question before I abandon this PR is the Gen 3 build check from scratch or from sstate?

gunnarx:
> One question before I abandon this PR is the Gen 3 build check from scratch or from sstate?

I commented on the ticket, but repeating for the benefit of other readers:  PR build checks reuse sstate for speed.  For complex cases I suggest setting up a dev branch and separate pipelines.  Not that it helps much b/c we need robust builds also with sstate in the end.

PR #88, opened 2017-05-23
Various init.sh update

gunnarx:
Not a really clean PR here since there might be somewhat unrelated changes.  I was going to push these directly in fact - but I'm concerned with one of the commits removal of official support targets might break a few builds that need the -force flag.  So I'll run it through the PR.


gunnarx:
Most build checks have finished. No issue noticed with init at least.  Unsupported hardwares have already been moved to other build pipelines so they're not affected. I guess it's ready to merge.

PR #93, opened 2017-06-14
Chromium multi-seat implementation

jaragunde:
This patch set adds the multi-seat implementation to the chromium build in GENIVI.

The multi-seat patch has already been modified to pull the ozone-wayland fork that implements [multi-seat](https://github.com/jaragunde/ozone-wayland/tree/wip/multi-seat).

Unfortunately, we must wait to merge this PR. It includes a downstream patch for wayland-ivi-extension to fix GDP-616. The patch must be modified when a new bugfix release of that software happens.

### DIFF
--- a/meta-genivi-dev/meta-genivi-dev/recipes-extended/chromium/chromium-wayland_%.bbappend
+++ b/meta-genivi-dev/meta-genivi-dev/recipes-extended/chromium/chromium-wayland_%.bbappend
@@ -28,7 +28,7 @@ COMPATIBLE_MACHINE_armv7ve = "(.*)"
 
 # Renesas workarounds
 
-COMPATIBLE_MACHINE_m3ulcb = "(.*)"
+COMPATIBLE_MACHINE_rcar-gen3 = "(.*)"
 
 # Apply same TUNE_FEATURES as in an armv7a build
 ARMFPABI_armv7ve = "${@bb.utils.contains('TUNE_FEATURES', 'callconvention-hard', 'arm_float_abi=hard', 'arm_float_abi=softfp', d)}"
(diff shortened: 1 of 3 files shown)


gunnarx:
Looking at jaragunde:chromium-multi-seat, it doesn't look right to me - it seems to sit on the old chromium branch before meta-genivi-dev was merged in?   I sent you an email this morning and I thought we agreed you wait until I said that the chromium branch is ready...

Fortunately it seems to rebase without problem on the new setup.  I will take some time to double check and possibly push an update which sits on top of the new chromium branch.  Other things prioritized this afternoon however.

jaragunde:
@gunnarx I found that the chromium branch in genivi-dev-platform was already merged with meta-genivi-dev (it contained the commit 8a5c559 that merges meta-genivi-dev in, then the old chromium commits on top of it) so I modified the paths in my commits and re-applied them. I thought you had already finished with the chromium branch :)

gunnarx:
It was pushed but only to let the CI system try it out. :)  It turns out that there was some bug - a few details missed.  I have pushed a new version to test the build again and hopefully it is in better shape.

I just noticed GENIVI IT took the Go.CD server down just now (I have requested a proper SSL cert so it's because of that).  I will see tomorrow if everything is good.  If the branch builds OK we can then update this PR and test that too.

gunnarx:
As you can see, after I rewrote the target branch for this PR (chromium), the PR now looks odd and needs a rebase.  But there are conflicts.  Since I just messed with this stuff it's maybe most efficient that I take a stab at fixing it @jaragunde so you don't need to waste time on it.

gunnarx:
OK, chromium branch seems OK now. I put this on top and pushed it. 

I also figured out what was looking strange on my side - my jaragunde remote was actually pointing to meta-genivi-dev so I was comparing the branch there named chromium-multi-seat... :)

Reviewing...

jaragunde:
Thanks @gunnarx, this is looking good. let's see the r-car and minnowboard builds.

gunnarx:
The builds on r-car and minnowboard look alright to me.  Merging.

PR #96, opened 2017-06-20
Enable DRM backend for qemux86-64 image

gunnarx:
It's the content of PR https://github.com/GENIVI/meta-genivi-dev/pull/34 but rewritten so that it applies to GDP after meta-genivi-dev subtree inclusion.

[GDP-166] QEMU GFX performance drop

Enabling DRM kernel modules and gallium support in mesa. This allows
weston to use the drm-backend, when -vga std/virtio is used on the
kvm host.


gunnarx:
During today's sprint plan we decided to merge even though testing has been somewhat lacking.  There's a separate ticket to improve instructions for running qemu in various scenarios.

gunnarx:
(By the way, all 5 build checks were fine before rebase - it's showing failure only because they were aborted now)

PR #98, opened 2017-06-26
local.conf: Apply compilation hardening for GDP

phongt:
Add extra security compile options to CFLAGS and LDFLAGS.
Some packages are in blacklist:
	+ dbus-c++ (no -pie -fpie flags)
	+ lttng-ust (no -pie -fpie flags)
	+ python-uinput (no -pie -fpie flags)
	+ python-storm (no -pie -fpie flags)
	+ persistence-administrator (no apply any options)

Build and tested only on QEMU x86-64

more details:
http://www.yoctoproject.org/docs/2.3/mega-manual/mega-manual.html#security-flags

Signed-off-by: Phong Tran <tranmanphong@gmail.com>

### DIFF
--- a/gdp-src-build/conf/templates/local.inc
+++ b/gdp-src-build/conf/templates/local.inc
@@ -33,3 +33,13 @@ BB_DISKMON_DIRS = "\
 # track the version of this file when it was generated. This can safely be ignored if
 # this doesn't mean anything to you.
 CONF_VERSION = "1"
+
+require conf/distro/include/security_flags.inc
+
+# Fixme: these pkgs are bronken while enable security_flags
+SECURITY_CFLAGS_pn-dbus-c++ = "${SECURITY_NO_PIE_CFLAGS}"
+SECURITY_CFLAGS_pn-lttng-ust = "${SECURITY_NO_PIE_CFLAGS}"
+SECURITY_CFLAGS_pn-python-uinput = "${SECURITY_NO_PIE_CFLAGS}"
+SECURITY_CFLAGS_pn-python-storm = "${SECURITY_NO_PIE_CFLAGS}"
+SECURITY_CFLAGS_pn-persistence-administrator = ""
+SECURITY_LDFLAGS_pn-persistence-administrator = ""


chbae:
This is great patch but we need to discuss to apply this. :)
Or we can optionally add this.

gunnarx:
Thanks @chbae.  Actually we already had some discussion on this previously so when phong mentioned it again I asked for a PR.

The class adds gcc flags -fstack-protector-strong, -fpie, -D_FORTIFY_SOURCE=2 to all compilation.
From a theoretical point of view it's a little strange that a distribution need to add some of these flags - it seems it would be better that the component themselves already had this by default in their Makefiles, and that the job of distros were to select only components that are secure, or get them fixed upstream.  But it is nowadays the norm for most Linux distros to add these compiler settings, as another tool to avoid something slipping through.

I think most of the time these settings get enabled on all final systems.  I've seen it in production builds.  So I prefer not to create any complexity to make this optional, because everything is already "optional" in some sense.  Anyone can easily replace the local.inc with their own, or modify it again in their developer.conf.    I also don't believe in any final product development that will not do some modification of distro settings for optimizing their system, whether the product is based on GENIVI Baseline/GDP, AGL or something else.  But our job is to select the most reasonable default and provide it as a starting point.   Compiling with these flags enabled seems reasonable, so if we can confirm there is no negative behavior, it can go in.

However, doesn't it belong better in the distro settings?  In that case it might be pushed into meta-ivi instead, which is where GDP gets its base distro definition at the moment.  Parts of the blacklist could be GDP specific however, so it creates a different split.

Let me know please @chbae - if you see some other risk.  Thanks.

phongt:
@gunnarx  

> However, doesn't it belong better in the distro settings? In that case it might be pushed into meta-ivi instead, which is where GDP gets its base distro definition at the moment. Parts of the blacklist could be GDP specific however, so it creates a different split.

I agree that the configuration should put to distro setting in meta-ivi [poky-ivi-systemd.conf](https://github.com/GENIVI/meta-ivi/blob/master/meta-ivi/conf/distro/poky-ivi-systemd.conf). More details information the packages in black list are provided by

1. meta-ivi: persistence-administrator
2. meta-genivi-dev: dbus-c++, python-uinput, python-storm, lttng-ust.

gunnarx:
I created a [ticket to include this in meta-ivi](https://at.projects.genivi.org/jira/browse/BASE-49).  If it is put into the meta-ivi distro configuration, after that GDP will still need to blacklist a few packages as shown in this patch, and that only applies for GDP.   But we should also report to those component projects to investigate why the flags cannot be used of course.

gunnarx:
I added a ticket in [https://at.projects.genivi.org/jira/browse/BASE-49] to ask if this will be taken into meta-ivi distro instead.

gunnarx:
Still waiting for the meta-ivi ticket to be processed (vacations).  But since it is open there and we need to adjust this patch anyway after meta-ivi is modified, I'll close this one for now.

PR #89, opened 2017-06-08
renesas-rcar-gen3: Rebase on Yocto BSP v2.19

slawr:
The Genivi 12 branch of the Renesas R-Car Gen 3 Genivi Yocto BSP has
been updated to use v2.19 of the upstream Yocto BSP. Update to this
newer version.

This has been sanity tested on the following boards:
M3 Starter Kit
M3 Salvator-X
H3 Salvator-X

Signed-off-by: Stephen Lawrence <stephen.lawrence@renesas.com>

I tried to use the git-submodule-bump script to create the commit but
could not get it to work as expected. If someone can show how it works
I can push a new commit if you wish.

PR #100, opened 2017-06-27
meta-ivi: Use wayland-ivi-extension v1.11.1

gunnarx:
Bumped meta-ivi to latest on 12.0 branch. This also means we can remove
the locally applied patch for a crashing bug which is fixed in in the
newer version of wayland-ivi-extension.  At the moment that patch is
applied on chromium feature branch only, so there's nothing (yet) to
remove on master branch, but I still want this commit on master.  The
patch removal will thus follow this in a separate commit.  I won't
squash them since it would just complicate merging the feature branch
later.

b0adb53 wayland-ivi-extension: update to release v1.11.1


gunnarx:
Merged #89 (Renesas BSP upgrade), rebased this one and pushed it ->one more build check.

PR #102, opened 2017-06-29
Bbappend restructure

gunnarx:
Warning:  Pushing it for build test - I expect several build failures, and will push updates accordingly.
(You don't need to review this until finished)

-----
Move .bbappends into "namespace" directories
    
All .bbappend (with one or two remaining exceptions) are moved to fully qualified subdirectory names indicating which layer they are extending/modifying.
    
This is according to previous analysis ([GDP-609](https://at.projects.genivi.org/jira/browse/GDP-609), which links to discussion material proposals, mailing-list discussion and the documented naming convention resulting from the consensus).
    
It may not be perfect, but it was the rough consensus and something we want to try for a while.
    
[GDP-609] Create a strategy for where to put stuff in the...


gunnarx:
No, not merged, this was a mistake. I have reset master to what it was.  New PR coming.

PR #105, opened 2017-07-04
Upgrade to Pyro

chbae:
[GDP-605] Upgrade base system to Yocto 2.3 / Pyro branch

PR #106, opened 2017-07-05
Pr 105 (Pyro) merged with bbrestructure

gunnarx:
Only pushing this for build tests and for @chbae to look at.  

I created a merge commit (see commit message for more details) where the new bbappend added in #105 
are also moved to new locations.  It's easy to miss extra files like patches and such, so I expect probably some build errors here.


chbae:
Thanks @gunnarx. I'll check and rebase #105

gunnarx:
@chbae -  yes, except that once this is properly fixed there is probably no more rebase necessary.  If you can do a better job on it, sure, but I did not find a good way, or the merge looked better for me at least.
I'm looking to fix the remaining build errors.

gunnarx:
The erlang build fail was just fixed and I amended the merge commit and pushed.

gunnarx:
Updated again.

gunnarx:
For simplicity the latest commits on master were put on top instead of below and I added some more fixes to the merge.  This allows us to compare to #105.  As it looks now, #105 will probably be the basis for the merge to master, but I want to support @chbae finding out what is different first, and why this builds correctly (or at least it did - we'll see).

I found some qt patches that had not been properly removed when solving the merge conflicts, and some
unnecessary parts in meta-raspberrypi-gdp.  That whole sublayer can be removed (extension remains under meta-raspberrypi name)

@chbae, if you want you can compare #105 and #106 now.  There are some diffs in audiomanager append - you will know better if you intended to have this or not - see:

meta-genivi-dev/meta-ivi/recipes-multimedia/audiomanager/audiomanagerplugins_7.5.bbappend

The other differences are because we have different naming structure -- I explained in #105 about it.

gunnarx:
Merged in chbae's latest qemu/mesa fix:
e9e9d16fcbe765a41 mesa: Fix build error for qemux86-64 
--> is 45f344b here

gunnarx:
Closing this since I don't think we have any use for it because #105 looks like it is in good shape.

PR #107, opened 2017-07-07
init.sh: Add a simple help to avoid locale errors

gunnarx:
Bitbake (python in fact) requires a terminal locale that supports UTF-8. Most of the time users will have that but not always.  Also in some simplified environments like CI-builds it might be missing.  For simplicity's sake I figured we can have init.sh set a locale like English/US/UTF-8 in case it needs to be done.

### DIFF
--- a/init.sh
+++ b/init.sh
@@ -128,8 +128,25 @@ function setupGitSubmodules() {
 }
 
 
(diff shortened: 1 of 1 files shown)


gunnarx:
Thanks for the review guys. Merging.

PR #108, opened 2017-07-07
Better source archive (ci-build script)

gunnarx:
Let the ci build script do the settings for source and license info instead of pipelines

Signed-off-by: Gunnar Andersson <gandersson@genivi.org>

PR #109, opened 2017-07-07
Create a file list for the image (ci-build script)

gunnarx:
ci-build.sh: Create a file list for the image
    
    For some situations it's useful to just quickly confirm if a file exists
    in the system. If publishing a small text file containing a file
    manifest then there's no need to download a large image or tarball to do
    such checks.
    
    ci-build.sh will now do that by default and it's possible to turn off
    by env var CREATE_ROOTFS_FILE_LIST = false
    
    Note: It was simply written for .tar.gz rootfs packages now - so this
    might need tweaking if we don't include a tarball in IMAGE_FSTYPES in
    some cases.
    
### DIFF
--- a/scripts/ci-build.sh
+++ b/scripts/ci-build.sh
@@ -284,11 +284,14 @@ if [[ "$BUILD_SDK" == "true" ]]; then
   bitbake meta-ide-support
 fi
 
+cd "$BASEDIR"
 rm -f logs.tar logs.tar.gz
 find gdp-src-build/tmp/work \( -name "*.log" -o -name "log.*" -o -name "run.*" \) -print0 | xargs -0 tar uf logs.tar || true
 gzip logs.tar || true
 
-cd "$BASEDIR"
+# Soften up the failure requirements here.  Maybe sometimes
+# some things can't be staged, which is probably ok.
+set +e
 rm -rf staging
 shopt -s nullglob
 stage_artifact mv gdp-src-build/tmp/deploy/licenses
@@ -300,6 +303,7 @@ stage_artifact mv logs.tar.gz
 stage_artifact cp gdp-src-build/buildhistory/images/*/glibc/genivi-dev-platform/files-in-image.txt
 stage_artifact mv gdp-src-build/buildhistory
 stage_artifact mv gdp-src-build/tmp/buildstats
+set -e
 
 # Environment contains alot of variables from Go.CD that specify the built
 # version/hash, and other metadata.  Let's store them for future reference.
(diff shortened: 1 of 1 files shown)


gunnarx:
Note that the first two commits are covered by #107 and #108.  It's quicker to compound them like this than to rebase this repeatedly once they are merged.

slawr:
Just a related comment, rather than a review, in case you are not aware. Yocto creates a rootfs file list as part of the manifest.

gunnarx:
Great.  I had a feeling...  Well if the file is always created anyway then no reason to not always include it in the artifacts.  Updated.

slawr:
I've never sanity tested the file list against the actual rootfs to prove they are the same.
I've also not tested when it gets updated to prove to myself exactly when the contents
get updated. Does it get completely wiped for example even if the build is unsuccessful.
That thought prompts me to do that sometime :)

That said, the build history folder contains various useful files and I personally archive it
alongside any build snapshot. Tolkien does the same and makes it available with the 
meta-ivi releases.

gunnarx:
Good point.  The script should archive the whole folder, or at least provide the option.  The builds are now failing in all kinds of weird ways.  I'm not sure what's going on but I should try to separate the commits.  Certainly the line append_local_conf 'INHERIT += "archiver"' of the first commit is the only one that could reasonably affect the build behavior...

gunnarx:
Isolated the commit and re-pushed.  Gotta get some answers on what's actually failing.

gunnarx:
OK, builds fine, maybe something fishy with the other commit that was included.  

Anyway, I'll update once more and include buildhistory and also buildstats in the artifacts.  That makes sense and it's not a lot of data in comparison to others.

scripts/ci-build.sh
```diff
@@ -296,7 +296,10 @@ stage_artifact mv gdp-src-build/tmp/deploy/licenses/genivi-dev-platform*/license
 stage_artifact mv gdp-src-build/tmp/deploy/sdk*
 stage_artifact cp gdp-src-build/tmp/deploy/images/*
 stage_artifact cp gdp-src-build/conf/*.conf
-stage_artifact cp logs.tar.gz
+stage_artifact mv logs.tar.gz
+stage_artifact cp gdp-src-build/tmp/buildhistory/images/*/glibc/genivi-dev-platform/files-in-image.txt
+stage_artifact mv gdp-src-build/tmp/buildhistory
```
chbae: Can this line is covered to upper line?

gunnarx: Yes, that is correct.  There will be two copies of files-in-image.txt in the result.  But the file is small and I wanted it to be easy to find at the top of the staging directory, and with a well-known path.  To get to the other files, you would look in the full build history directory.

slawr: Agree that many might not find the files buried in buildhistory. Especially as you have to go through several folder levels to find any files. Just a small suggestion if you return to this in the future - realising you probably have more important things to do than forever finesse this PR. Maybe consider having a symlink to the files file in buildhistory rather than a copy.

gunnarx:
Merging.  If there are other improvements that can be made to the script we can always add more later.

PR #110, opened 2017-07-10
Compile chromium for Dragonboard

gunnarx:
Just out of interest, modified the compatible machine and inserted the same gypi file as R-Car to make this compile. It seems to compile OK but is completely untested.  This PR is not intended to be merged right away but mostly to get it into some kind of testing/evaluation/further work queue.


chbae:
@gunnarx As our policy, we need to move to meta-genivi-dev/meta-browser/recipes-browser/...

gunnarx:
Thanks @chbae, you are right.

gunnarx:
This is still waiting for testing.  Sorry for the delay, it is kind of low priority.

PR #114, opened 2017-07-25
Sandbox HMI App with bubblewrap PoC

DunderRoffe:


gunnarx:
I wonder if the container could have more namespaces enabled, like --unshare-net, but it's a good first step.
Looks good to me, let's merge it.

DunderRoffe:
I have pushed this to a feature branch. Lets have a discussion before merging to master :)

PR #117, opened 2017-07-30
meta-oic: meta-iot-web: Update to current 1.2.1 branch

rzr:
* meta-oic: update for iotivity-1.2.1
* meta-iot-web has been also updated for iotivity-mode 1.2.1

Signed-off-by: Philippe Coval <philippe.coval@osg.samsung.com>


gunnarx:
+1.  Thanks RzR

rzr:
you're welcome this will prevent a gcc-7 issue

gunnarx:
Looks good to me.  This is just a layer update.  Perhaps @DunderRoffe and @frznlogic want to chime in, or just merge?  I think we're not testing this a whole lot but since it builds and we trust @rzr to do this right ;-)

rzr:
Nothing much that just a version upgrade (1.2.1) packages groups are same, other benefits are compilation enhancement and also it will support demos examples shared in meta-ocf-automotive as introduced at bottom of this page:
https://blogs.s-osg.org/ocf-automotive-connected-car-demos-ces-2017/

More resources at:
http://wiki.iotivity.org/automotive

cc: @iamsanjeev

gunnarx:
It's more about our preferred format of commit messages.  

Note that the git-submodule-bump script is not perfect because it handles only one submodule at a time.  So if these two submodule bumps depend on each other they should be one commit of course, but then you might just need to manually add the shortlogs of both modules.  You could run the script twice and then squash the commits after. If they're independent then of course just make it two commits.

No rush on this though, thanks for looking at it when you have time again.

rzr:
Hi, 
Is there anything wrong with build jobs ?

gunnarx:
>  Is there anything wrong with build jobs ?

Unfortunately we (I...) broke master last week and many PR builds failed because of it.
As soon as we saw it, it waas fixed in aa7d013.  I will rebase this PR on tip of master and push it back to trigger a rebuild.

PR #125, opened 2017-08-21
Use only git:// links for submodule setup

gunnarx:
We are experiencing really strange network problems in CI at the moment.  Some indication that it happens over HTTP protocol made me want to try this, to see if it has any positive effect.  

(In any case it shouldn't hurt to be consistent).

We have preferred to use HTTPS for most other git URLs, but most of the submodules were git:// already, so I figured there's nothing to lose from this.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "meta-ivi"]
 	path = meta-ivi
-	url = http://github.com/genivi/meta-ivi.git
+	url = git://github.com/GENIVI/meta-ivi.git
 [submodule "meta-qt5"]
 	path = meta-qt5
 	url = git://github.com/meta-qt5/meta-qt5.git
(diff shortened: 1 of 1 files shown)


gunnarx:
It had no effect and there's something wrong with the PR build plugin.  I can't imagine what it might be because this seems to have just broken spontaneously.  Will have to look at it deeper...

rzr:
http[s] is more proxy friendly, and could help for yocto adoption.

gunnarx:
@slawr, @rzr  I think you both might have missed the point, and my initial comment.

@rzr - Yes, the first part of the last comment I made:

_"We have preferred to use HTTPS for most other git URLs, but most of the submodules were git:// already, so I figured there's nothing to lose from this."_

was regarding a policy decision we took before to use  HTTP(s) in all bitbake recipes.  The reason behind it
was indeed corporate proxy issues.

This was just investigation into a problem and it had to be sent as a PR since it was investigating PR check builds.  

If you read it, you can all logically see that the PR, even if accepted, can have no negative impact on the proxy issue since you have to be able to clone *all* submodules to successfully build.  And git:// is already used!

> So my vote would be to use http where possible

@slawr  We're in full agreement, "where possible".  

First I wanted to understand why GitHub seemed to from time to time send strange responses on HTTP(s) requests which failed the PR builds.

gunnarx:
Pushed an update to try once more to investigate the build behavior a bit more using consistent URLs.

gunnarx:
(By the way, I did not change the commit comment and it already outlines all the background regarding proxies etc.)

gunnarx:
PR builds now seem OK for some reason.  But they are failing due to broken master which is being fixed by #131

frznlogic:
The master has been fixed in #131, we need to rebase and build it again to get status of the PR.

gunnarx:
Just to clear @frznlogic I'm NOT actually asking this to be merged (see comments above).  I'd still like to see the build result.

gunnarx:
This builds successfully, but also other PRs do now, without this change.  Guess we close this for now and see if the problem reappears or not.

PR #129, opened 2017-08-29
Pyro integration 

frznlogic:


frznlogic:
The pyro branches have been tested and seem to work on all but the renesas-rcar-gen3 platform. Separate issue is opened for that platform. Rebasing and merging this PR.

PR #130, opened 2017-09-01
genivi-dev-platform.bb: Fix randomized metadata

frznlogic:
I'd like to see if anyone has a better solution for this commit, ie do we need to keep the $DATE, if so, how do we add it?

### DIFF
--- a/meta-genivi-dev/meta-genivi-dev/recipes-dev-platform/images/genivi-dev-platform.bb
+++ b/meta-genivi-dev/meta-genivi-dev/recipes-dev-platform/images/genivi-dev-platform.bb
@@ -8,7 +8,7 @@ DESCRIPTION = "GENIVI Development Platform image which includes currently a simp
 HMI and three proof-of-concepts (PoC's) and demos. The GENIVI Browser PoC, \
 the Fuel-Stop-Adviser (FSA) PoC and the GENIVI AudioManager (AM) Demo."
 
-PV = "1.3+snapshot-${DATE}"
+PV = "1.3+snapshot"
 
 BOOT_SPACE_rpi = "40960"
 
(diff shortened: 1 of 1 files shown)


meta-genivi-dev/meta-genivi-dev/recipes-dev-platform/images/genivi-dev-platform.bb
```diff
@@ -8,7 +8,7 @@ DESCRIPTION = "GENIVI Development Platform image which includes currently a simp
 HMI and three proof-of-concepts (PoC's) and demos. The GENIVI Browser PoC, \
 the Fuel-Stop-Adviser (FSA) PoC and the GENIVI AudioManager (AM) Demo."
 
-PV = "1.3+snapshot-${DATE}"
+PV = "1.3+snapshot"
```
chbae: Why don't you use PV to GDP version" like PV = "12"

gunnarx: @chbae We could of course have the major number in there, but I'm not sure how much it will help during the development phase.   We have a pretty long development cycle between a release until the next release, and during all that time there is nothing but "master", and some identifier.  I would be unsure exactly when to switch the PV number, to be honest, (unless of course it's at the very end when the release is ready, i.e. at the same time as we do tagging).

We should really use the commit hash of the gdp project if you don't like the date, @frznlogic

frznlogic:
Rebased and testing again. 

We had a discussion on whether to accept this or not right now but decided to think about a proper scheme to use until tomorrow. One problem exists, I have no really good way of getting a nice git hash or something into the version string, it would greatly improve ability to see what we're actually running.

frznlogic:
The tests failed due to a CI agent acting up. Re-running them now to see results.

gunnarx:
> I have no really good way of getting a nice git hash or something into the version string

Two ideas:
1. The go.cd build agent will make sure there is an environment variable defined containing the hash.  But that's only valid in the go environment.
2. If you could execute a script here then just let it go to the root of the project and run 
"git rev-parse HEAD"

gunnarx:
I'm trying out variable expansion using the git command wrapped in python.

gunnarx:
We have some necessary discussion and decisions to take here, (and also iterate towards a solution that works).  I prefer to do that in the ticket, so please refer to [GDP-743](https://at.projects.genivi.org/jira/browse/GDP-743)

gunnarx:
Let's just go with "13.0" for now.

frznlogic:
Doing this straight on master, we know what we want and how to solve it.

PR #131, opened 2017-09-01
nativesdk: Remove dnf support

frznlogic:
The dependency chain for the nativesdk-dnf is currently broken. DNF is currently not used
and/or supported by the genivi-dev-platform anyways, so removing the dependency.

### DIFF
--- a/meta-genivi-dev/poky/meta/recipes-core/packagegroups/nativesdk-packagegroup-sdk-host.bbappend
+++ b/meta-genivi-dev/poky/meta/recipes-core/packagegroups/nativesdk-packagegroup-sdk-host.bbappend
@@ -1,0 +1,2 @@
+RDEPENDS_${PN}_remove = "nativesdk-dnf"
+


gunnarx:
OK for me if this solves the problem.  Hmm, that's odd - apparently the PR builds now started running again without giving up early on some http/network/Go-plugin/whatever error.  So let's see if the builds succeed, I guess.

chbae:
This cause is ./meta-flatpak/recipes-support/gpgme/gpgme_%.bbappend

- BBCLASSEXTEND = "native"

Need to contribute.

frznlogic:
@chbae thanks for the input! I've moved the bbappend for now. Do you know how to remove something from the BBCLASSEXTEND? I failed at using the _remove syntax.

chbae:
@frznlogic  How did you use _remove syntax? I think that BBCLASSEXTEND_remove will work.

frznlogic:
@chbae nope, it didn't unfortunately. I'm sorry for doing this, but I'll rebase and merge this commit just so we can get a working baseline again. If we want to fix it some other way, I'm open for suggestions. 

Basically though, what this does is turn off a feature we're not interested in (the native support for DNF package management), as a bonus it also removes a build error.

gunnarx:
Oscar and I also talked about this offline.  I support a quick fix for this because master was broken and that we want to fix as soon as possible of course.  Happy to accept any other proposals for changes after this.

PR #133, opened 2017-09-07
Fix build against gcc 7.1.1

zeenix:
On my Fedora 26 host, these patches are needed to get GDP build successfully.

### DIFF
--- a/meta-genivi-dev/meta-genivi-dev/recipes-extended/dbus-c++/dbus-c++-native.bb
+++ b/meta-genivi-dev/meta-genivi-dev/recipes-extended/dbus-c++/dbus-c++-native.bb
@@ -3,4 +3,5 @@ inherit native
 
 DEPENDS = "glib-2.0-native dbus-native expat-native"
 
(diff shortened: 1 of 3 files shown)


gunnarx:
It's worth documenting for the future that we discussed that the patches would probably be better "upstream", but no one seems to want to take care of the maintenance of dbus-c++ at the moment, as far as we are aware (please let us know if you think otherwise).

Alt 1: @tom--pollard We're fetching dbus-c++ from git.baserock.org but that's just legacy reasons I think.  Would someone at Codethink receive the patch - if so how do you want it delivered?  (Or you just take it here)

Alt 2: 
- Put a copy of dbus-c++ on github.com/GENIVI and apply the patch

Alt 3:
- Here's another fork: https://github.com/Pelagicore/dbus-cplusplus
but I'm told this is deprecated from Pelagicore point of view, recommending glib_mm instead, so it won't be a reliable upstream.

With this said, I still +1 the PR for now.

frznlogic:
I agree to the points made by gunnar. The preferred solution is probably to swap to a better maintained dbus implementation. Failing that we need to at least maintain our codebase

PR #134, opened 2017-09-12
Adding vehicle_signal_interface_2.0.0.bb

frznlogic:

### DIFF
--- a/meta-genivi-dev/meta-genivi-dev/recipes-dev-platform/images/genivi-dev-platform.bb
+++ b/meta-genivi-dev/meta-genivi-dev/recipes-dev-platform/images/genivi-dev-platform.bb
@@ -32,6 +32,7 @@ IMAGE_INSTALL_append = " \
     packagegroup-nodejs-runtime \
     packagegroup-gdp-qt5 \
     packagegroup-gdp-rvi \
+    packagegroup-gdp-vsi \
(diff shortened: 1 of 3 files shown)


gunnarx:
Build checks:  Some OK, some fail.  Failures due to erlang checksum mismatch (not on some agents that have the old file cached) - ref #136.

meta-genivi-dev/meta-genivi-dev/recipes-dev-platform/images/genivi-dev-platform.bb
```diff
@@ -32,6 +32,7 @@ IMAGE_INSTALL_append = " \
     packagegroup-nodejs-runtime \
     packagegroup-gdp-qt5 \
     packagegroup-gdp-rvi \
+    packagegroup-gdp-vsi \
```
gunnarx: I think we might have a discussion/redesign regarding package groups.  There's a risk of having almost one per component, which isn't the purpose. 

The group you created here could serve as the "vehicle-signaling-subsystem" group for any related components.

We might rather call it -vss  although to be clear vss stands for v.s.specification, not v.s.subsystem.  Nonetheless, vss might be a useful name for this group.

gunnarx:
Rebased. Rebuilding.

PR #136, opened 2017-09-12
Fix meta-erlang properly (& use temporary fork)

gunnarx:
The proper fix is to avoid the use of tarballs from GitHub altogether.

I've sent a PR upstream.  Using a fork in the meantime.  

(!) Untested commit, trying build.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -25,7 +25,7 @@
 	url = git://github.com/GENIVI/meta-rvi.git
 [submodule "meta-erlang"]
 	path = meta-erlang
-	url = git://github.com/joaohf/meta-erlang.git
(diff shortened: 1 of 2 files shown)


gunnarx:
This now builds OTP v19 and fails on applying patches that probably don't apply to the new version.

gunnarx:
Reference to the [meta-erlang PR](https://github.com/joaohf/meta-erlang/pull/8).  It has been merged and maintainer says: "seem to be correct".

gunnarx:
Considering this obsolete now because of #137.  Later on we should upgrade to latest meta-erlang, but the urgent build issue must be fixed first.

PR #137, opened 2017-09-13
Erlang hotfix keeping version 18.2.3

gunnarx:
New attempt.  Refer to #136 for more details.

The difference is this branches meta-erlang from the version we had, and does therefore not update to the latest tip of master.  Thus, the OTP version is held back at 18.2.3 which is what we had.  Hopefully this compiles.

### DIFF
--- a/gdp-src-build/conf/templates/local.inc
+++ b/gdp-src-build/conf/templates/local.inc
@@ -10,6 +10,9 @@ LICENSE_FLAGS_WHITELIST  += "commercial"
 
 #INCOMPATIBLE_LICENSE ?= "GPLv3"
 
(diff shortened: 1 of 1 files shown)


gunnarx:
Hmm.  It is now building 18.2.3 instead of 19, but the same patch application is still failing the same as for #136.  I don't get it.

gunnarx:
What Oscar pushed before the weekend built just fine. We can squash the contradictory commits (e.g. layer forked to gunnarx, then to frznlogic, and now back on upstream again).  I pushed new commits - they are also rebased on latest master.   Review this cleaned up version for mergeability (but we should merge ASAP since it's in fact a broken master branch)   

I did my best to make a clean history while also amending to Oscar's patches so he is still credited for the changes.

PR #142, opened 2017-09-25
Remove nativesdk-packagegroup-sdk-host.bbappend

chbae:
meta-flatpak is updated to fix dnf parsing problem.
So nativesdk-packagegroup-sdk-host.bbappend is not required anymore.

Signed-off-by: Changhyeok Bae <changhyeok.bae@gmail.com>


gunnarx:
Nice one @chbae!

Might be just in time, because I think I ran into an "dnf missing" on one of the SDK build.  I don't see it now, but it could be hidden.  So this should bring back dnf and avoid that problem maybe.

PR #144, opened 2017-09-25
Run services as dedicated system users

zeenix:
This is the PR from my efforts on https://at.projects.genivi.org/jira/browse/GDP-722 . I'm afraid I didn't get too far but I think it should give a good start for anyone who wishes to complete this goal. As said in the commit log, the latest commit doesn't actually work for me but it could be something with my host as some folks said that it really should just work.


frznlogic:
A general question, have you tried if the swm stuff will work? I'm guessing the service is used to update installed software, can we still do this with the updated users?

zeenix:
@frznlogic yes, at least the services can successfully launch. That's why we need write access to that /usr/lib/xxx dir.

gunnarx:
Discussed in team today.  We're holding on this one since we don't feel confident that the software installation parts will work without further testing furst.

gunnarx:
Rebased on master.

gunnarx:
Rebased on master.

mirzak:
Superseded by #178

PR #148, opened 2017-10-02
Feature sdl pyro

frznlogic:
Testing the feature sdl core for integration.


gunnarx:
Build check:  Failing on all non-Intel platforms because of this file: linux-intel_%.**bbappend** which means Yocto will look for a linux-intel_%.**bb** to append to, which won't exist on the non-intel builds.

Actually the layer [naming convention](https://at.projects.genivi.org/wiki/x/w4Xk) addresses the problem in this particular case.

Instead of:

meta-genivi-dev/**meta-genivi-dev**/.../linux-intel.bbappend
it should be
meta-genivi-dev/**meta-intel**/.../linux-intel.bbappend

and that sub-layer will only be included on the Intel based builds.

(Similarly, sublayer for poky appends should be used for the other append).

meta-genivi-dev/meta-genivi-dev/recipes-dev-platform/images/genivi-dev-platform.bb
```diff
@@ -38,6 +38,7 @@ IMAGE_INSTALL_append = " \
     packagegroup-gdp-rvi \
     packagegroup-gdp-vsi \
     packagegroup-gdp-sdl \
+    packagegroup-gdp-vsi \
```
gunnarx: I was wondering about this commit since it seems unrelated to the function.  The commit seems already merged on master.   Somehow you doubled it - the same commit information is found twice in the commit history of this (PR) branch.

gunnarx:
@frznlogic I think we can close this for now?  There are separate build pipelines building the feature branch,
and the branch has also been updated since this PR.

PR #150, opened 2017-10-04
FSA update 

gunnarx:
Received this not as a PR but as a patch set via email, from @blochm

We have still some design discussions on this code but we can progress it later.  For example, I'd like to try to remove patches that are applied by the recipe.  In several cases it is patching GENIVI repositories, so there should be little issue with just fixing the code upstream instead.

Considering this as a feature branch and hold on merging for now.  It's not tested (or reviewed) enough to go into 13.0 release.  Perhaps 13.1.


gunnarx:
Rebased on master.

gunnarx:
Rebased on master

gunnarx:
Rebased for a new build check.

gunnarx:
Rebased on master again.  Just because...

gunnarx:
Does not build without gypsy which is nowadays dropped from meta-openembedded
because bluez4 is dropped.  Before that, also this happened acbe0268202a9e09f95ed088c7a4abec4b106cc4...

gunnarx:
This will not be updated.

PR #155, opened 2017-11-01
Add script to update meta-ivi's dependencies

gunnarx:
When meta-ivi updates, it defines the compatible commits/versions of
dependent layers such as poky and meta-openembedded in its README.

After updating meta-ivi, running this script will update the layers in
GDP to the versions proposed by the baseline.  Sometimes we may choose
to deviate from it, but it is a starting point that is likely to be
compatible.

### DIFF
--- a/scripts/update_baseline_dependency_layers.sh
+++ b/scripts/update_baseline_dependency_layers.sh
@@ -1,0 +1,43 @@
+#!/bin/sh
+
+# Part of genivi-dev-platform
+# (C) Gunnar Andersson
+# License: MPLv2
+
+# Update poky, etc. to the versions required by the current meta-ivi baseline
+
+fail() {
+  echo "Something failed.  You might need to clean the layer somehow (see above).  Giving up."
+  exit 1
+}
+
+update_version() {
+  layer=$1
+  cd meta-ivi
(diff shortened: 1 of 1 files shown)


scripts/update_baseline_dependency_layers.sh
```diff
@@ -0,0 +1,43 @@
...
+
+fail() {
+  echo "Something failed.  You might need to clean the layer somehow (see above).  Giving up."
+  exit 1
+}
+
+update_version() {
+  layer=$1
+  cd meta-ivi
+  v=$(scripts/get_layer_info_from_README.sh $layer revision)
```
frznlogic: I can't find the get_layer_info_from_README.sh script anywhere, is it made available through a different PR?

frznlogic: Ah,  thanks! That makes sense I guess ;)

gunnarx: Yes, especially since the line before it is "cd meta-ivi" ;)

frznlogic: Rub it in my face, why dont you ;)

PR #157, opened 2017-11-01
ci-build: Restore user.name/email

gunnarx:
The ci-build script would clobber the user's user.name and user.email
settings in git, because that makes no difference in a throw-away CI
environment.  It is required (in some cases, such as when git is used as
patch tool) that there is SOME setting for user/email in the CI build.
But since ci-build.sh could be useful to run also locally, the original
settings must be restored when the script ends.

### DIFF
--- a/scripts/ci-build.sh
+++ b/scripts/ci-build.sh
@@ -117,7 +117,6 @@ stage_artifact() {
   done
 }
 
-
 # Layer version overrides:
 # Find and evaluate environment variables of type
 
@@ -193,7 +192,6 @@ print_layer_overrides() {
   done
 }
 
-
 # FIXME: Asking git to list submodules would be better
 LAYERS="
 meta-browser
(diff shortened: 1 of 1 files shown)


scripts/ci-build.sh
```diff
@@ -376,3 +378,16 @@ fi
...
+  git config --unset user.name
+else
+  git config user.name "$olduser"
+fi
+if [ -z "$oldemail" ] ; then
+  git config --unset user.email
+else
+  git config user.email "$oldemail"
+fi
+
```
frznlogic: A bit over the top perhaps, but I think this kind of "restore" functionality might be nice to add into a function which is run at the end of the script, as well as having some signal handlers setup to restore in case of failure and/or user signalling the script to stop. For example:

```
#!/bin/bash

foo() {
  echo bar
  exit
}

trap "foo" SIGINT SIGTERM

doStuff...

foo
```

gunnarx: Updated

frznlogic:
LGTM

PR #158, opened 2017-11-01
CI: enable override of one particular layer version

gunnarx:
Enables building a version of GDP parent project, with a particular layer exchanged for a different version.
Launching various on-demand build checks and other CI scenarios is easily 
done by setting new environment variables, per layer.

There are a few more commits here, with cleanup etc.

### DIFF
--- a/scripts/ci-build.sh
+++ b/scripts/ci-build.sh
@@ -6,8 +6,10 @@
 # User warning!
 
 # A casual user will need to read the script to understand what the first
-# parameter shall be to make it run:
+# parameter shall be to make it actually run:
 #
+
+# Fail script unless you give the magic value:
 if [[ "$1" != CI_FLAG ]]; then
 
   cat <<EOT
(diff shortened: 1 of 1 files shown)


scripts/ci-build.sh
```diff
@@ -6,22 +6,26 @@
 # User warning!
 
 # A casual user will need to read the script to understand what the first
-# parameter shall be to make it run:
+# parameter shall be to make it actually run:
```
frznlogic: Not sure that I like this kind of comment, if the script is meant for casual users, this should be documented at the top of the script IMHO. Just my .2$.

scripts/ci-build.sh
```diff
@@ -110,6 +117,103 @@ stage_artifact() {
...
+      value=$(deref "$var_name")
+      if [ -n "$value" ] ; then
+        echo "$var_name = $value"
+      fi
+    done
+  done
+}
+
+
+# FIXME: Asking git to list submodules would be better
```
frznlogic: Agreed! meta-sdl for example is missing in there....

gunnarx: This is a PR for master, so it is correct to not include meta-sdl.  But yes, it has a high risk of being forgotten when SDL is merged.

scripts/ci-build.sh
gunnarx: Maybe the wording is strange, but the intended meaning is rather to say that it's **not** intended for casual users.  A casual user that attempts to run it will be forced to read it first...

The point here is that the script is not that user friendly, but that's not the comment but the code in lines 13 to 31, and that was put there deliberately.  The reason is  1) It's a little bit  dangerous script to use as a casual user, so therefore it really should not be run without reading and understanding it first,  and 2) Supporting CI build is the priority task.  (It's all written in 13-31 so I won't repeat more).  The comment is there to explain the code.

PR #159, opened 2017-11-01
meta-renesas is now renesas-rcar-gen3

gunnarx:
It must be done like this, right @slawr?

### DIFF
--- a/scripts/ci-build.sh
+++ b/scripts/ci-build.sh
@@ -248,7 +248,8 @@ git submodule status
 # Deal with special setup, copy binary drivers etc.
 set -x
 if [[ "$TARGET" == "r-car-m3-starter-kit" ]];  then
-  cd meta-renesas
+  echo "Copying binary graphics drivers for $TARGET"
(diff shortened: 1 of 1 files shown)


slawr:
Is this related to R-Car Gen 3? If yes, then the commit is wrong. If you can explain a bit about what you are trying to do I may be able to more specific.

For Pyro the R-Car Gen 3 YBSP layer is meta-renesas.
The adaptions to configure meta-renesas for Genivi-13 are in meta-ivi-renesas.

renesas-rcar-gen3 was the name of the combined Yocto BSP and Genivi support for R-Car Gen 3 prior to Pyro/Genivi-13, i.e. Genivi 10-12.

See my [blog](https://at.projects.genivi.org/wiki/pages/viewpage.action?pageId=16024340) on the subject

gunnarx:
Yeah, I think this was an old outgoing change I found and I must have got it reversed.  
I couldn't quite figure out how the builds were succeeding before, but that's because it was already correct.
So... please ignore :)

PR #162, opened 2017-11-20
ci-build.sh: Add GENIVI MIRROR + PREMIRROR option

gunnarx:
Adding environment variables to define MIRRORS and PREMIRRORS variables. Only one URL for each is supported here.  PREMIRROR is by default empty, and MIRROR is by default set to a GENIVI hosted mirror.  (MIRRORS is used by Yocto as a fallback if all other source fetch options failed)

For PREMIRRORS it is normally more convenient to use the own-mirrors BB class, but since MIRRORS need to be explicitly defined, I used the same format for PREMIRRORS.

It's quite arbitrary but I chose to apply the PREMIRROR definition first (prepend) and MIRROR definition last (append), in case there were any settings already.  Seems somewhat logical that this is the desired
behavior.

NOTE that keeping the GENIVI mirror automatically up to date is not yet implemented, but the content is up to date with 13.0 release.  In any case :  

1. Only what _is_ available there will be fetched.
2. Since it is used for MIRROR, not PREMIRROR, it's only used as a fallback if local DL_DIR, and upstream URLs fail.  => It is no worse than without the mirror.

NOTE2 : I tested that the syntax of appending to local conf seems correct.  
Fetching from the mirror also seems to work fine.

### DIFF
--- a/scripts/ci-build.sh
+++ b/scripts/ci-build.sh
@@ -247,6 +247,8 @@ define_with_default BUILD_SDK false
 define_with_default COPY_LICENSES false
 define_with_default LAYER_ARCHIVE false
 define_with_default CREATE_RELEASE_DIR false
+define_with_default MIRROR "https://docs.projects.genivi.org/releases/yocto_mirror"
+define_with_default PREMIRROR ""  # By default none (but we have the shared DL_DIR)
 define_with_default RM_WORK false
 define_with_default REUSE_STANDARD_DL_DIR true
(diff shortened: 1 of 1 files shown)


frznlogic:
Looks good to me. Do you plan to implement mirror content updating for the genivi mirror?

gunnarx:
Yes, "plan to".  There's quite a few projects in the pipe, though.

PR #165, opened 2017-12-06
Update meta-ivi & configure hardening flags

gunnarx:
Nitpick: the commit comment says it is updating to master.  In reality I pushed branch wip_hardening_flags to GENIVI/meta-ivi so the commit can be referenced from GDP while it's still only an open PR in meta-ivi.
If we merge that without additional changes, the commit history and "latest master" will be correct.  Otherwise I can rewrite this commit to match.

(see commit comment for log)

### DIFF
--- a/gdp-src-build/conf/templates/local.inc
+++ b/gdp-src-build/conf/templates/local.inc
@@ -13,6 +13,15 @@ LICENSE_FLAGS_WHITELIST  += "commercial"
 PREFERRED_VERSION_erlang-native = "18.2.3"
 PREFERRED_VERSION_erlang = "18.2.3"
 
+# Blacklist for compilation hardening flags.
+# The poky-ivi-systemd distro file in meta-ivi includes security_flags.inc
+# Some packages apparently fail if (some) flags are applied, so they are
+# updated here to not apply them, or apply them partially.  This lists the
+# packages that are in GDP only, not in meta-ivi.
+SECURITY_CFLAGS_pn-dbus-c++ = "${SECURITY_NO_PIE_CFLAGS}"
+SECURITY_CFLAGS_pn-python-uinput = "${SECURITY_NO_PIE_CFLAGS}"
+SECURITY_CFLAGS_pn-python-storm = "${SECURITY_NO_PIE_CFLAGS}"
+
 # Disk Space Monitoring during the build
 #
 # Monitor the disk space during the build. If there is less that 1GB of space or less


gunnarx:
meta-ivi PR reference https://github.com/GENIVI/meta-ivi/pull/60

gunnarx:
Pushed an update with a longer commit comment for "Blacklist some packages from compiler hardening flags"

frznlogic:
Do you mind if we wait until the chrome pr is pulled in first?

gunnarx:
I don't mind.  PR merge order is always whatever team thinks is appropriate.
Rebased for new build check.

gunnarx:
We'll just take this as part of the rocko update since meta-gplv2 is updated there.

phongt:
please note that the gettext fixing is only merged in [master](https://git.yoctoproject.org/cgit/cgit.cgi/meta-gplv2/log/)

gunnarx:
Rebased this on latest origin/master, and new build checks.  I expect it might still fail however since the PR is built on master (GDP 13) but requires related changes in baseline that are only on 14 branch now (if I recall correctly).  Anyway, I'm not aware of anything more that really needs to be done on this - it should be OK once merged toward baseline/GDP 14 status.  Will confirm.

gunnarx:
A lot of churn right now that we need to get through but overall this PR looks quite alright.

So, to comment on the build failures after the last rebase.  For R-Car it's an issue that the CI is not supporting both versions of SGX binary drivers at the same time.  CI agents have been updated to the one compatible with GDP 14, and therefore the master (GDP 13) that the PR is applied to fails in CI builds.  Supporting more than one is solvable of course, but it's not done that way now.  Anyway not a fault of this PR.

For Dragonboard there's a link failure in chromium build.  This *could* presumably be affected by the hardening flags, or not.  I think we'll know only after merging #177 and applying this on top.

I'd say this PR needs to go on top of #177 only after that has been merged.

gunnarx:
By the way, this PR updates meta-ivi in d63f92a, but also #177 and friends update to even later versions.  So presumably that commit should be omitted when this is based on top of GDP 14.

gunnarx:
This is way obsolete.  For example meta-ivi update in 28028f8  (on master branch) brings meta-ivi to tag P-0.1.  It supersedes this PR, except for the hardening flags that were apparently merged a bit later.  Next update to P-1.0 should bring in all of it.  Let's redo it.

PR #166, opened 2017-12-07
Adding h3ulcb as a compatible machine for chromium-wayland

rkollataj:
chromium-wayland: Adding h3ulcb as a compatible machine.

Building GDP image for r-car-h3-starter-kit target is currently not possible as h3ulcb is not set as compatible for chromium-wayland. This commit is fixing the issue.

Signed-off-by: Remigiusz Kołłątaj <remigiusz.kollataj@mobica.com>

### DIFF
--- a/meta-genivi-dev/meta-genivi-dev/recipes-extended/chromium/chromium-wayland_%.bbappend
+++ b/meta-genivi-dev/meta-genivi-dev/recipes-extended/chromium/chromium-wayland_%.bbappend
@@ -29,6 +29,7 @@ COMPATIBLE_MACHINE_armv7ve = "(.*)"
 # Renesas workarounds
 
 COMPATIBLE_MACHINE_m3ulcb = "(.*)"
+COMPATIBLE_MACHINE_h3ulcb = "(.*)"
 
 # Apply same TUNE_FEATURES as in an armv7a build
 ARMFPABI_armv7ve = "${@bb.utils.contains('TUNE_FEATURES', 'callconvention-hard', 'arm_float_abi=hard', 'arm_float_abi=softfp', d)}"


gunnarx:
Thanks @rkollataj Patch looks good to me.  There's a build failure but that might be just an issue with this branch.  We might rebase this patch onto another branch and in any case currently @frznlogic is in the middle of working on merging Chromium feature branch to Master.

gunnarx:
Oh, there's also the possibility that we should instead switch the MACHINE override to "rcar-gen3" which covers all boards.  The most general choice is usually preferred, but depends on if we feel other boards need to be tested with this feature first.

frznlogic:
I would probably prefer it against either feature_sdl_pyro and/or feature_chromium and/or master (after feature_sdl_pyro has been merged). We can cherry-pick it over though.

frznlogic:
I've copied the change over to a separate branch.. Probably unnecessary, but I felt more comfortable doing it this way. A new PR vs master has been created at PR 167, closing this one as chromium is now on master. I will rework the names of the VSI branches etc FYI.

rkollataj:
Ok, thanks for the info. I was struggling a bit to figure out which branch I should choose. What is a preferred workflow in such situations? Should I create JIRA ticket first to discuss where to put potential fixes?

frznlogic:
Hi,

Yeah I can't blame you, the chromium integration has been quite messy, especially since a lot of other features depended on it :). 

What we winded up with was master -> feature_chromium(_pyro) -> feature_sdl(_pyro) -> feature_chromium_sdl_vsi_pyro

Now that feature_sdl_pyro is integrated to master we should only have feature_chromium_sdl_vsi_pyro left, which should be renamed to indicate that it's only specific for the VSI integration really.

In general,  I think it's ok to put the PR up and we can always see if there's any comments about it here. If you are uncertain, there's the #automotive irc channel, and the mailing lists to ask for further guidance/help etc on, see https://lists.genivi.org/mailman/listinfo/genivi-projects. I'd probably go for genivi-projects for this type of questions.

gunnarx:
Adding:
- That any published feature branch had a build error was unexpected, but yeah cleanup was ongoing. 
- Whichever one you had chosen, it would have needed to be merged to other branches.
+1 on all the rest.

slawr:
I'll add to the thanks to @rkollataj for sending the patch.

As an example the MACHINEOVERRIDES for a M3 Starter Kit build with the R-Car Gen 3 YBSP are:
"rcar-gen3:r8a7796:cortexa57-cortexa53:m3ulcb:aarch64:ulcb"
So as Gunnar noted the best generalisation for Gen 3 SoC based boards is rcar-gen3.

When I took a quick look at the diff I wondered what the workaround the existing comment preceding it was referring to. I'll see if I can remind myself and comment which might be best.

slawr:
Taking a look at the wider recipe I would say that yes this should be generalised to "rcar-gen3" as I believe among other things the statement is referring to the backported gcc patch in that recipe. That gcc version is used by all the Renesas boards supported in the YBSP. So lines 31 and 32 should be squashed into a single statement with rcar-gen3 as the modifier.

As an aside that gcc fix can likely be dropped when we move to the next big YBSP update end of January, which moves to gcc v7.x.

When looking at the git blame that brought in the original m3ulcb change I noticed something. The igalia [commit ](https://github.com/GENIVI/genivi-dev-platform/commit/e58dd634f91ec2cfa13ed14f9cd8bce8e6705ab6) from @hferreiro that brought it in was specific to the M3 Starter Kit board and uses m3ulcb as the subdirectory name for the file folder. That should be generalised as well else it will not be applied to the H3 boards or M3 Salvator-X.

PR #169, opened 2017-12-08
meta-gplv2: Bump meta-gplv2 because of known issues in gettext

frznlogic:
c7f97f1 groff: Drop local getopt implementation in favor of the one from glibc
08afd87 groff: Make it build with compile time hardening enabled
4bed302 elfutils: Make it build with GCC 7 and compile time hardening enabled
9b1d684 gnupg: specify tar path explicitly
d866801 gmp_4.2.1: prevent calls to mpn_add_nc() if HAVE_NATIVE_mpn_sub_nc is false
4fb2d41 mc: disable Ctrl-g key and Ctrl-\ key
5396115 Revert "mc: Drop from meta-gplv2 as incompatible with newer ncurses versions"
cc59f9f Revert "layer.conf: Remove mc from packagegroup-core-full-cmdline-utils"
73a8f9d layer.conf: Remove mc from packagegroup-core-full-cmdline-utils
ec80d2c mc: Drop from meta-gplv2 as incompatible with newer ncurses versions


gunnarx:
Since we apparently don't have an urgent build error, why not hold on this and update once gettext patch is merged upstream.  (See https://github.com/GENIVI/meta-ivi/pull/60)

gunnarx:
meta-ivi is about to merge significant updates for rocko branch.  After that we can update to latest meta-ivi, and dependencies like meta-gplv2 will follow.  Since we know this will happen (driven by #165) I think this PR can be closed.

PR #170, opened 2017-12-08
chromium-wayland: Depend on wayland-ivi-extension

frznlogic:
Occasionally we had a compile error where the header file ilm_common.h
was not found during compilation of chromium-wayland.  This header comes
from wayland-ivi-extension.  The error seems to have gone away (or is
depending on some random build order?).  In any case I can't imagine
anything else than that the version of chromium-wayland we build should
DEPEND on wayland-ivi-extension.


frznlogic:
Dragonboard failure is expected. Need to pull in the dragonboard patch from PR 139

PR #171, opened 2018-01-24
fix build error: meta-erlang

spa46:
Fetching meta-erlang is fixed and pull request is sent.
(https://github.com/GENIVI/meta-genivi-dev/pull/97)

### DIFF
--- a/gdp-src-build/conf/templates/bblayers.inc
+++ b/gdp-src-build/conf/templates/bblayers.inc
@@ -17,6 +17,7 @@ BBLAYERS ?= " \
   ${TOPDIR}/../meta-openembedded/meta-gnome \
   ${TOPDIR}/../meta-openembedded/meta-python \
   ${TOPDIR}/../meta-qt5 \
+  ${TOPDIR}/../meta-genivi-dev/meta-erlang \
   ${TOPDIR}/../meta-genivi-dev/meta-genivi-dev \
   ${TOPDIR}/../meta-oic \
   ${TOPDIR}/../meta-iot-web \


gunnarx:
My apologies for the delay...  Merged the m-g-d PR now.  The submodule sha1 hash changed because we always rebase before merge, so I updated and pushed back a new commit to your branch here.  Hopefully g-d-p repo can now build OK.  Restarting build checks.

I guess the BSP has been reorganized for R-Car since 12.x, and I think that's causing the CI build error there.  (The build pipeline is set up to build 13 and later).  So I'll ignore that for now.  If anyone is working on R-Car with the old 12.x branch, please feel free to check the build manually.

slawr:
> I guess the BSP has been reorganized for R-Car since 12.x, and I think that's causing the CI build error there. (The build pipeline is set up to build 13 and later). So I'll ignore that for now. If anyone is working on R-Car with the old 12.x branch, please feel free to check the build manually.

I took a look at the failure. You are correct. GDP-12 points at https://github.com/slawr/renesas-rcar-gen3/tree/genivi-12/meta-rcar-gen3 for the R-Car Gen 3 Genivi Yocto BSP, which is correct and is still available to support just this kind of work. From Genivi-13/GDP-13 the git rep became https://github.com/slawr/meta-renesas to align with the upstream YBSP layer name and to allow it to be used in various community work. It seems the pipeline assumes the Genivi-13 YBSP root path name to find the gfx/mmp package copy script (meta-renesas rather than renesas-rcar-gen3) which results in the error.

gunnarx:
OK, with that explanation it can be merged.  Please check if everything is OK (there is some risk the CI system used cached sources and therefore will not confirm the source fetching error)

PR #172, opened 2018-02-27
Update meta-ivi baseline + Audiomanager

gunnarx:
Pushing a PR for build test.  For explanation, see commit comment.

PR #173, opened 2018-03-07
Update to baseline 14 & rocko branch

gunnarx:
This is continuing  #172.  I switched branches so I had to recreate the PR.


gunnarx:
Added #174 (chromium updates) on top of this, + one more fix.

slawr:
Some notes on R-Car Gen 3:

I have a build of P-0.1 Genivi 14 Yocto Baseline on R-Car Gen 3 building locally based on upstream YBSP v3.4 and my meta-ivi-renesas. YBSP v3.4 brings various changes such as adoption of Poky meta gcc v7.x so I would not spend time trying to get this PR working with YBSP v2.23.1 that GDP master currently points at.

I assume (hope) GDP would build as well but that is yet to be proved with the latest changes in this PR. I am leaving early today but will try early next week. YBSP v3.4 introduces two new meta layers to the build so I want to check whether they are needed for GDP before sending commits to update the R-Car Gen 3 layers in GDP.

gunnarx:
Noted, will debug the basic stuff on the other targets for now, and then we can update R-Car.

slawr:
Sounds like a plan - should be straight forward Monday to try. If not, or just because you want to build for SK, I can provide local.conf and bblayers for local build if you wish. It's straight forward.

Actually I'll check if the additional layers are needed and then send commits anyway as a WIP. No need to keep them private.

gunnarx:
We have a lot of open PRs now.  I want to now consider this one obsolete and close it.  

Most of my comomits are in @chbae's new #173, except for one or two commits that were re-done independently by @chbae and in a different order than I did.   But in any case we based later combined PRs on @chbae's #173, so we can get rid of this one now.

PR #174, opened 2018-03-07
WIP: bump Chromium version

jaragunde:
This pull request bumps the version of Chromium from 53 to 63.

Instead of using the latest version of the Chromium recipes from the meta-browser layer, it uses a different one from [meta-lgsvl-browser](https://github.com/lgsvl/meta-lgsvl-browser), who maintain a newer version of Chromium with the Ozone/Wayland backend that we were using in previous releases. The version of Chromium-Wayland provided by meta-browser doesn't support the IVI shell so I would like to skip it for this release and reconsider for the next one.

This couple of patches have been tested with Raspberry Pi 3 and they should work with 2. Builds for R-Car gen 3 are known to be crashing at this point.

TODO:
* fix R-Car gen 3 build
* test minnowboard
* test dragonboard
* rebase multi-seat patches (if time permits)

PR #176, opened 2018-04-06
Rocko combined

gunnarx:
Yet another PR... 
* combines #175 and [chbae/#2](https://github.com/chbae/genivi-dev-platform/pull/2)
* Fixes sign-off on the meta-intel commit (thanks @slawr)

Sure it's a bit messy right now with all the redundant PRs, but it's an effective way to get some test builds on these different combinations, so it should be all good in the end.

If this one goes well, the next step is to add back Chromium again, taken from #174, which should then work on R-Car (but as we can see there seem to be a few issues on other platforms)


slawr:
@gunnarx thanks for the merge.

Just a reminder, for R-Car Gen 3 YBSP v3.6.0 you want the gfx/mmp package [R-Car Gen3 Evaluation Software Package for Linux
for Yocto v3.6.0 and Wayland 1.13 /Weston 2.0](https://www.renesas.com/en-us/solutions/automotive/rcar-demoboard-2.html) on the CI.

Which consists of the userland and kernelland downloads (R-Car_Gen3_Series_Evaluation_Software_Package_for_Linux-20180326.zip and R-Car_Gen3_Series_Evaluation_Software_Package_of_Linux_Drivers-20180326.zip) 

Without the update I expect the build will still succeed but if you want to actually use it you will likely hit problems.

gunnarx:
Update: added a CI script fix dad30c5f0598f46b38d2c67ea7eb0aa4d993ef59 fixing a bug that might be responsible for the R-Car failed build in CI

gunnarx:
> Just a reminder, for R-Car Gen 3 YBSP v3.6.0 you want the gfx/mmp package R-Car Gen3 Evaluation Software Package for Linux for Yocto v3.6.0 and Wayland 1.13 /Weston 2.0 on the CI.

For completeness - note that the CI agents that support Renesas builds now have this version of the drivers (but unfortunately not the old ones, which means GDP 13 can no longer be built in CI for Renesas. Soon that should not be a concern any more)

This PR could probably be closed in favor of #177, but I think we need to debug the final issues and merge and then we can clean up redundant PRs.

slawr:
> For completeness - note that the CI agents that support Renesas builds now have this version of the drivers (but unfortunately not the old ones, which means GDP 13 can no longer be built in CI for Renesas. Soon that should not be a concern any more)

Just a suggestion but should the need arise I would use the SGX directory var to handle this. To handle GDP 13 simply point the SGX directory var of the CI script at the directory for the v2.23.1 gfx packages as a pre-requisite for the pipeline.

gunnarx:
Added on one commit: Mirza's SDK fix that removes qtwebkit also from SDK (already removed from image).

gunnarx:
The new Chromium seems to work OK, so close this in favor of #177

PR #178, opened 2018-05-03
[WIP] GDP-722: List all used services and make sure they have a dedicated user

mirzak:
Supersedes #144 

Verified that the following services at least start:

- package_manager.service
- aktualizr.service
- ambd.service
- lifecycle_manager.service
- module_loader_ecu1.service
- partition_manager.service
- software_loading_manager.service

I do not yet know how to run-time test them and to what extent they are to be tested. I am pretty sure that "swm" services will not work without additional changes. E.g running "rpm/dpkg" requires elevated privileges which the current user (swm) does not have.

HMI is left out, since we need to change Weston first to launch with the "display" group.

Still putting this out there for initial review.

### DIFF
--- a/meta-genivi-dev/meta-genivi-dev/recipes-sota/genivi-swm/files/0001-common-settings-update-DB_URL-to-var-run-swm-swlm.sq.patch
+++ b/meta-genivi-dev/meta-genivi-dev/recipes-sota/genivi-swm/files/0001-common-settings-update-DB_URL-to-var-run-swm-swlm.sq.patch
@@ -1,0 +1,41 @@
+From 69ee3d50bd26ff6708327558d0e7933136631c6c Mon Sep 17 00:00:00 2001
+From: Mirza Krak <mirza.krak@endian.se>
+Date: Thu, 3 May 2018 12:42:52 +0200
+Subject: [PATCH 1/1] common: settings: update DB_URL to
+ /var/run/swm/swlm.sqlite
+
+/var/run is root owned, and if we run the software_loading_module as a
+non-root user we will have issues with maintaining permissions, even if
+we can create an initial file with the correct permissions. If the
+process attempts to re-create the database it will need to remove the
+current file and fail when trying to create a new file in a root owned
+directory
+
+By moving the swl.sqlite file to /var/run/swm, we can control the
+permission of the "swm" directory allowing processes running under the
+same user/owner as the directory to modify its content to the fullest.
+
+Signed-off-by: Mirza Krak <mirza.krak@endian.se>
+
+Upstream: pending
+
+---
+ common/settings.py | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+diff --git a/common/settings.py b/common/settings.py
+index 8cf44efe6b85..19af8b30e9ec 100755
+--- a/common/settings.py
++++ b/common/settings.py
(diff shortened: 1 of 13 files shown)


mirzak:
@gunnarx, seems that the builds are failing due to low availability of storage space.

```
13:20:28.182 WARNING: The free space of /var/cache/yocto/downloads (rootfs) is running low (0.993GB left)
13:20:28.183 ERROR: No new tasks can be executed since the disk space monitor action is "STOPTASKS"!
13:20:28.184 WARNING: The free space of /var/cache/yocto/sstate (rootfs) is running low (0.993GB left)
13:20:28.184 ERROR: No new tasks can be executed since the disk space monitor action is "STOPTASKS"!
```

meta-genivi-dev/meta-genivi-dev/recipes-sota/genivi-swm/files/0001-common-settings-update-DB_URL-to-var-run-swm-swlm.sq.patch
```diff
@@ -0,0 +1,41 @@
+From 69ee3d50bd26ff6708327558d0e7933136631c6c Mon Sep 17 00:00:00 2001
```
gunnarx: Regarding swm we ought to change the original project source and drop the patch.  But since this is going to be reworked anyway, maybe there's no point in optimizing now.

mirzak: My idea was:

- Merge here
- Create pull-request upstream
- Drop patch here once upstream was updated

Mostly because I was unsure who has control (maintainership) of the upstream repo and avoiding "blocks".

gunnarx:
> 13:20:28.182 WARNING: The free space of /var/cache/yocto/downloads (rootfs) is running low (0.993GB left)
By the way, that's not the main build area storage, but the SSD that holds sstate and downloads cache only.  For the two agents go-agent1 and go-agent2 that use this I have moved the downloads to the slower disk -- keeping the sstate there is more important for speed, so this should happen more seldom now.

meta-genivi-dev/meta-genivi-dev/recipes-sota/genivi-swm/files/0001-common-settings-update-DB_URL-to-var-run-swm-swlm.sq.patch
gunnarx: Yeah, we are thinking the same here - I just noted it for the future.

gunnarx:
Let's ignore the R-Car build fail which is due to mismatched graphics drivers. (They were updated to match Renesas R-Car Gen 3 Yocto BSP v3.6.0).  In the next BSP update we said we would fix this so that multiple versions are buildable at the same time without any fuss.

gunnarx:
> E.g running "rpm/dpkg" requires elevated privileges which the current user (swm) does not have.

Since the SOTA part will be redone we'll take this as-is now, even if it fails.
,

mirzak:
I also plan on pushing HMI + weston updates (run as non-root) to this branch.

gunnarx:
Rebased...

gunnarx:
This will not be updated.

PR #183, opened 2018-05-28
README: update current release and fix broken link

mirzak:
Current release is actually GDP 13 and the download page has been renamed so update the link.

Signed-off-by: Mirza Krak <mirza.krak@endian.se>


gunnarx:
For the future maybe we should simplify the README and not have information that gets out of date.
There are (short) links in the Wiki that are stable even when a page gets renamed.
For now, let's take this quickfix though.

PR #186, opened 2018-06-01
GDP-611 - Prepare system for libostree

mirzak:
Basically moved everything that is installed in `/opt` to `/usr/share`, apps are installed to /usr/apps instead of /opt.

Have done a basic analysis of what files are being accessed during a running image, note this is basically "idle" state and hard to test all paths in all applications without deep insight in each component.

Here is a 60 seconds recording:

```
root@qemux86-64:~# inotifywatch -v -t 60 -e modify -e close_write -e create -e delete -r /
Establishing watches...
Setting up watch(es) on /
OK, / is now being watched.
Total of 11291 watches.
Finished establishing watches, now collecting statistics.
Will listen for events for 60 seconds.
total  modify  close_write  create  delete  filename
36     18      18           0       0       /run/
36     18      18           0       0       /var/volatile/log/
30     30      0            0       0       /usr/share/rvi_core/log/lager/
30     30      0            0       0       /run/log/journal/e889500dab0e4187a07b39181af55991/
24     6       6            6       6       /run/systemd/journal/streams/
24     0       24           0       0       /dev/
24     12      12           0       0       /sys/fs/cgroup/cpu,cpuacct/system.slice/
12     6       6            0       0       /sys/fs/cgroup/devices/system.slice/system-serial\x2dgetty.slice/
12     6       6            0       0       /sys/fs/cgroup/blkio/system.slice/
12     6       6            0       0       /sys/fs/cgroup/memory/system.slice/
12     0       0            6       6       /sys/fs/cgroup/systemd/system.slice/system-serial\x2dgetty.slice/
12     12      0            0       0       /var/volatile/log/smartdevicelink/
4      2       2            0       0       /sys/fs/cgroup/systemd/system.slice/system-serial\x2dgetty.slice/serial-getty@ttyS1.service/
4      2       2            0       0       /sys/fs/cgroup/systemd/system.slice/system-serial\x2dgetty.slice/serial-getty@ttyS1.service/
4      2       2            0       0       /sys/fs/cgroup/systemd/system.slice/system-serial\x2dgetty.slice/serial-getty@ttyS1.service/
4      2       2            0       0       /sys/fs/cgroup/systemd/system.slice/system-serial\x2dgetty.slice/serial-getty@ttyS1.service/
4      2       2            0       0       /sys/fs/cgroup/systemd/system.slice/system-serial\x2dgetty.slice/serial-getty@ttyS1.service/
2      1       1            0       0       /sys/fs/cgroup/systemd/system.slice/system-serial\x2dgetty.slice/serial-getty@ttyS1.service/
1      1       0            0       0       /dev/pts/
```
The only problem with above is that `rvi` is writing to `/usr/share` which is a `read-only` area when booting with libostree. Worst case rvi crashes, best case rvi fails gracefully to write logs.

This is "mergable" to me and further optimizations can be done later when found.


gunnarx:
> The only problem with above is that rvi is writing to /usr/share which is a read-only area when booting with libostree. Worst case rvi crashes, best case rvi fails gracefully to write logs.

Thanks for analyzing that.  I just made a simple search/replace of paths in rvi, and expected there might be some issues.  Looking for /usr/share/rvi_core/log/*  and moving that to /var should be the right approach.

meta-genivi-dev/meta-rvi/recipes-connectivity/rvi/files/rvi.service
```diff
@@ -12,8 +12,8 @@ Environment="RVI_BACKEND=38.101.164.230"
 
 StandardOutput=journal
 StandardError=journal
-ExecStart=/bin/sh -c "RVI_MYIP=$(/sbin/ip route | /usr/bin/awk '/default/ { print $3 }') CONFIG=/etc/opt/rvi/rvi.config RVI_BACKEND=38.101.164.230 /opt/rvi_core/rvi_ctl -c /etc/opt/rvi/rvi.config console"
-ExecStop=/opt/rvi_core/rvi_ctl stop
+ExecStart=/bin/sh -c "RVI_MYIP=$(/sbin/ip route | /usr/bin/awk '/default/ { print $3 }') CONFIG=/etc/opt/rvi/rvi.config RVI_BACKEND=38.101.164.230 /usr/share/rvi_core/rvi_ctl -c /etc/opt/rvi/rvi.config console"
```
gunnarx: CONFIG= should change.  In the RVI fix I changed /etc/opt/rvi/* to /etc/rvi/*

meta-genivi-dev/meta-genivi-dev/recipes-dev-hmi/genivi-dev-platform-hmi/gdp-new-hmi.bb
```diff
@@ -1,34 +1,43 @@
...
 do_install_append() {
-        install -d ${D}${libdir}/systemd/user
-        install -m 0444 ${WORKDIR}/gdp-new-hmi.service \
-                        ${D}${libdir}/systemd/user
-        mkdir -p ${D}/home/root/.config/systemd/user/default.target.wants/gdp-new-hmi.service
-	ln -sf /usr/lib/systemd/user/gdp-new-hmi.service ${D}/home/root/.config/systemd/user/default.target.wants/gdp-new-hmi.service
+    install -d ${D}${systemd_user_unitdir}
+    install -p -D ${WORKDIR}/gdp-new-hmi.service ${D}${systemd_user_unitdir}/gdp-new-hmi.service
+    install -d ${D}${sysconfdir}/systemd/user/default.target.wants
+    ln -sf ${systemd_user_unitdir}/gdp-new-hmi.service ${D}${sysconfdir}/systemd/user/default.target.wants
```
gunnarx: You're installing the symlink to a different location (/etc).  I know because I just made the same change :)  and I agree... :)
I think this fixes the dependencies again so the startup is correct.

mirzak: I believe that I am just mimicking what `systemd --user enable <service>` would do if executed on target.

But I will check again.

I saw your change and left a comment on it :)

gunnarx:
Just documenting our discussion for 3d62626, regarding standard location for installing "apps":

> - application executables are put in /usr/share/gdp-apps
>   - No real "standard" exists for application binaries

* /usr/share shall according to [FHS](https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html) have "architecture independent files", which binaries and libs are not.  QML-files and graphics fit the bill however.
* The problem with FHS / Unix standard way, is of course that splitting things into bin, lib, share "correctly" breaks apart app files that we would probably rather prefer to keep within one directory.
* possibly /usr/libexec/<appname> might be the most correct alternative.  It is for "internal" binaries (i.e. not typically exposed to the user).  There could be an entrypoint (e.g. symlink) added in /usr/bin that refers to these internal binaries.  But on the other hand, if Desktop files (or even .service files for that matter) are the real entry point then I suppose you could argue that these are "internal".
* If there is no apps standard, and if we will break the standards that exist, why not skip share and go with:  /usr/gdp-apps/<appdir> ?
* By the way, I think .../apps/ is better than .../gdp-apps/, since there are no other kind of apps here anyway.  I don't think GDP needs to be the kind of system that keeps including its own name in every single location possible, which just makes things unique for no good reason.  (Yes, I'm talking about _that other system_  ;-)
* Finally, maybe we should ask: WWFPD?  (What Would Flatpak Do? :-) :-)

I guess I'm still ready to approve the apps location, to move forward.  Don't know if there are other opinions?

gunnarx:
Added comment on rvi.  Should be squashed with the preceding commit.
If you're cleaning I think you can squash the two meta-oic commits also?

mirzak:
Rebased:

- squashed the RVI changes
- squashed the meta-oic append layer changes
- updated install path for apps to /usr/apps
- tagged all commits with JIRA ticket

gunnarx:
A few nitpicks to cleanup at later time regarding installation location etc. 
Looks good for now.

mirzak:
Missing RPi3,dragonboard and minnowboard build, hmm.

@gunnarx, what trick do you use to re-trigger builds?

gunnarx:
> @gunnarx, what trick do you use to re-trigger builds?

1. Typically I make some trivial amend / rebase and push again.
2. If they _were_ built and you want to "repeat the same build" for a particular one, then it can be done in Go.CD's web interface.  Because all PRs are built by the same pipeline there are some corner cases when that can't be done, such as if there was a new PR in between and when the pipeline configuration was changed.  In that case, option 1 again.

mirzak:
Restarted the tests. All GREEN now.

gunnarx:
continued on [os_tree](https://github.com/GENIVI/genivi-dev-platform/tree/os_tree) branch (and [rocko_os_tree](https://github.com/GENIVI/genivi-dev-platform/tree/rocko_os_tree))

PR #187, opened 2018-06-04
GDP-613: integrate meta-updater

mirzak:
To start an libostree enabled QEMU currently I have used the script available in `meta-updater`. With the following args:

    genivi-dev-platform/meta-updater/scripts$ ./run-qemu-ota genivi-dev-platform \
         --no-gui --force-kvm \
         --dir ../../gdp-src-build/tmp/deploy/images/ \
         --machine qemux86-64

Above results in following args to `qemu-system-x86-64`:

```
qemu-system-x86_64 \
        -bios ${IMAGE_DIR}/qemux86-64/u-boot-qemux86-64.rom \
        -drive file=${IMAGE_DIR}/qemux86-64/genivi-dev-platform-qemux86-64-20180601130742.otaimg,if=ide,format=raw,snapshot=on \
        -serial tcp:127.0.0.1:8990,server,nowait \
        -m 1G \
        -usb -device usb-tablet \
        -show-cursor \
        -vga std \
        -net user,hostfwd=tcp:0.0.0.0:2222-:22,restrict=off \
        -net nic,macaddr=ca:fe:83:81:c3:a7 \
        -nographic \
        -enable-kvm
```

The biggest difference with how we normally do it is the `-bios` and `-drive` arguments. 

libostree requires a boot-loader hence the `-bios` argument, In this case it will start U-boot and that will take care of the rest. 

Why boot-loader you might ask?

libostree requires a specific bootflow, and that is:

- boot to ramdisk (initrd) (takes argument of which deployment to use from boot-loader)
- ramdisk (initrd) sets up deployment and uses `chroot` to "boot" in to a specific deployment. Check [1] for more details.

libostree generates a specific disk image to be used (`otaimg`), this is a full disk image which sets up the root file system according to libostree requirements (moves everything to /usr, (from /lib and /bin etc..)) and this has to be used when booting a system to get a functional libostree configuration. And this is the updated `-drive` argument.

[1]. https://github.com/advancedtelematic/meta-updater/blob/rocko/recipes-sota/ostree-initrd/files/init.sh


mirzak:
Will soon push raspberrypi integration well, will be easier to discuss layout once that is done.

gunnarx:
Really good documentation.  Might be worth moving to the Wiki since PRs tend to be lost and forgotten.

mirzak:
> Really good documentation. Might be worth moving to the Wiki since PRs tend to be lost and forgotten.

Definitely will move this to a wiki page along with instruction on how to actually do an update.

mirzak:
Updated commits with ticket tag

mirzak:
Dropped raspberry-pi meta-updater support, will create a separate pull-request for that. There where some issues that where exposed in the CI environment that needs to be fixed and trying to avoid it to be a blocker for this pull-request which adds at least qemu support.

mirzak:
Should go in after #186

mirzak:
@gunnarx, this one is missing builds for some reason. Probably because we overloaded the builders today.

Re-trigger?

mirzak:
Restarting build and test again, last run gave an error on acceptance tests,  [here is the log](https://go.genivi.org/go/tab/build/detail/GDP-Yocto-QEMU_x86-64-GITHUB-PR/366/test/1/start_qemu_and_run_test)

It failed on `test_systemd_failed` while waiting for systemd to "settle" meaning:

"Block until systemd has finished activating any units being activated, or until two minutes has elapsed."

Above is probably related to the weston/HMI restarting which are causing instability in our tests.

mirzak:
On second though, I belive that the acceptance tests exposed a real problem. It seems that the service causing the problems is:

```
22:55:14.398 AssertionError: False is not true : Timed out waiting for systemd to settle:
22:55:14.398 UNIT              LOAD   ACTIVE     SUB          DESCRIPTION          
22:55:14.398 aktualizr.service loaded activating auto-restart Aktualizr SOTA Client
```

Which kinda makes sense since we do not boot the libostree image when running tests, which could probably cause some "wierd" behavior as the SOTA client might be expecting it. Will see if it is the same test result on this run, and this probably needs further investigation.

slawr:
If its any help SOTA has long been "noisy" with restarting systemd services. They don't start properly, due to system not being fully in place, so then they restart continuously. Gunnar knows more. I could not quite understand why they were not there, but not set to run by default along with a note "enabled foo when you actually want to do something meaningful with a fuller system"

gunnarx:
#186 and #187 are somewhat independent, but one is preparing for the other.  They both stem from master as they are written now.  I have combined the two in sequence (rebased 187 on top of 186) and placed that on a [os_tree](https://github.com/GENIVI/genivi-dev-platform/tree/os_tree) branch for easier build and test.   (Note, of course, only some hardware targets are supported so far.)

gunnarx:
> If its any help SOTA has long been "noisy" with restarting systemd services. They don't start properly, due to system not being fully in place, so then they restart continuously. Gunnar knows more.

Not that much help :) because most of what runs now is brand new because this uses an up to date aktualizer, and also a lot of that proof of concept software was [removed](
https://github.com/GENIVI/genivi-dev-platform/commit/a1d941073e7bb76e56c2abd900a4feb0cab7125d).  
and as Mirza said OS-tree boot needs to be included in test framework.  I'm think Mirza is on the case...

slawr:
>Not that much help :) because most of what runs now is brand new because this uses an up to date aktualizer, and also a lot of that proof of concept software was removed. and as Mirza said OS-tree boot needs to be included in test framework. I'm think Mirza is on the case...

Oh ok. Yeah ignore my comment then Mirza - it was just an off the cuff comment in case it helped :) Glad to hear its being cleaned up.

mirzak:
Yeah I was already planing including ostree in CI, https://at.projects.genivi.org/jira/projects/GDP/issues/GDP-794?filter=allopenissues.

This just shows that it probably needs to be part of this pull request to satisfy tests. Can not merge with tests failing :).

mirzak:
Glad at least that the tests are being useful, even if it is a quite small test suite it is valuable.

gunnarx:
> We should probably close this and #186 then. @gunnarx ?

I'll close them soon.  Maybe a little more conversation will happen, and might be useful for triggering some builds still.

gunnarx:
continued on [os_tree](https://github.com/GENIVI/genivi-dev-platform/tree/os_tree) branch (and [rocko_os_tree](https://github.com/GENIVI/genivi-dev-platform/tree/rocko_os_tree))

PR #189, opened 2018-06-06
templates: local.inc: remove package_rpm

mirzak:
RPM was our primary package manager which means that the
do_rootfs would use dnf to install all the required packages.

Currently RPM has performance issues when running inside a
docker container, which our builds normally do on our build
servers but it is also quite common to run this on developer
machines as well.

The performance issues are rapported and fixed in upstream RPM,
https://bugzilla.redhat.com/show_bug.cgi?id=1537564.

But since we do not rely on using RPM we simply remove it instead
of trying to backport and maintain a patchset for a couple of
Yocto release. Currently only poky/master has the "fixed" RPM
package.

[GDP-793] rpm in docker is slowing down builds

Signed-off-by: Mirza Krak <mirza.krak@endian.se>

### DIFF
--- a/gdp-src-build/conf/templates/local.inc
+++ b/gdp-src-build/conf/templates/local.inc
@@ -2,7 +2,7 @@
 
 DISTRO ?= "poky-ivi-systemd"
 USE_GSTREAMER_1_00 ?= "1"
-PACKAGE_CLASSES ?= "package_rpm package_ipk"
+PACKAGE_CLASSES ?= "package_ipk"
 EXTRA_IMAGE_FEATURES ?= "debug-tweaks"
 USER_CLASSES ?= "buildstats image-mklibs image-prelink"
 PATCHRESOLVE ?= "noop"
--- a/gdp-src-build/conf/templates/renesas-rcar-gen3.local.inc
+++ b/gdp-src-build/conf/templates/renesas-rcar-gen3.local.inc
@@ -19,8 +19,6 @@ SOC_FAMILY ?= "r8a7796"
 # Linaro GCC
 GCCVERSION = "linaro-5.3"
 
-PACKAGE_CLASSES = "package_rpm"
-
 # add the static lib to SDK toolchain
 SDKIMAGE_FEATURES_append = " staticdev-pkgs"
 


mirzak:
> I went ahead and amended the patch - check it ^^

Looks good.

> I'm good to merge if the build is well tested.

It is tested as it can be I would say. I have had rpm builds disabled on my build machine since day one. 

Testing clean build is definitely interesting but not a blocker for a merge IMO.

slawr:
The M3 SK sanity build failed because the gfx/mmp copy failed. That made me wonder if the CI script update in PR 190 (update to YBSP v3.7.0) caused an issue but it shouldn't. That PR was merged to the rocko branch and created a new directory for the YBSP v3.7.0 gfx/mmp drivers.

This PR should be using the earlier YBSP and it's gfx/mmp drivers in /var/go/sgx_bin_gen3/ ..

gunnarx:
> "20180326" is the pkgs for (rocko supporting) YBSP v3.6.0. I assume it was updated as part of the earlier work on rocko GDP.

Yeah, it's been updated all along and therefore master has been failing for a long time, thought you had noticed. This is why we updated the method now.  Just hang on until everything gets merged and gets caught up.

I don't understand a word of the rest, Stephen, but I'm sure it's fine :)  I would not over-analyze at this point in time.  Just relax and we'll be fine :)

Also ignoring that this comment ended on this PR which is only about RPM...?  
So based on that I will not know what "the M3 SK sanity build" actually builds.  Again, I don't think there is a problem.  Just one step at a time here.

slawr:
@gunnarx thought about this a little more. In part I probably wasn't too worried about the fact that GDP master CI build was broken for R-Car because the focus was on the Rocko update and that was ok.

However since then @mirzak has come on board and a bunch of new work is being done and merged on GDP master branch. The CI master branch issue means that presumably none of it is being sanity built on R-Car which makes me nervous.

Unless we will merge rocko very soon can we fix the CI master branch issue so sanity builds occur on R-Car again? The steps to fix it are just to replace the contents of /var/go/sgx_bin_gen3/ with the click through gfx/mmp zips (suffixed with "20170904") for YBSP v2.23 from [here](https://www.renesas.com/en-us/solutions/automotive/rcar-demoboard-2.html).

PR #190, opened 2018-06-07
rocko: Update meta-ivi layers. Update R-Car Gen 3 to Yocto BSP v3.7.0

slawr:
@gunnarx this contains two work streams but as these need to be merged into the rocko working branch I have not split them into separate PRs. Felt like it will just add a bunch of work to CI that is not needed. Rocko should be built with both anyway.

The two streams are:
1. Update to meta-ivi release 14.0.0 layer commits.
Whilst doing the Yocto BSP v3.7.0 update I noticed that GDP was still using the old P-0.1 release layers specified in the meta-ivi README.md. So there are commits to bring meta-oe, poky and meta-gplv2 up to date.

2. Update R-Car Gen 3 support to Yocto BSP v3.7.0 ([GDP-789](https://at.projects.genivi.org/jira/browse/GDP-789))
Update the Renesas R-Car Gen 3 board support to use Yocto BSP v3.7.0.

I would really like to see the BSP v3.7.0 update pulled into the rocko working branch to avoid having to repeat stability testing once the current general GDP HMI issues are fixed.

Warning: As discussed in this weeks BIT call the CI will fail the check build on R-Car until the CI is updated to contain the correct gfx/mmp packages for Yocto BSP v3.7.0. See [GDP-790](https://at.projects.genivi.org/jira/browse/GDP-790) for what needs to be done.

### DIFF
--- a/scripts/ci-build.sh
+++ b/scripts/ci-build.sh
@@ -40,7 +40,7 @@ fi
 AGENT_STANDARD_DL_DIR="/var/cache/yocto/downloads"
 AGENT_STANDARD_SSTATE_DIR="/var/cache/yocto/sstate"
 AGENT_STANDARD_SGX_LOCATION="/var/go/sgx_bin"
-AGENT_STANDARD_SGX_GEN3_LOCATION="/var/go/sgx_bin_gen3"
+AGENT_STANDARD_SGX_GEN3_LOCATION="/var/go/rcar-gen3/gfx-mmp_ybsp-370_20180423"
 
 # ---- Helper functions ----
 
@@ -471,8 +471,9 @@ fi
 
 # Deal with special setup, copy binary drivers etc.
 if [[ "$TARGET" == "r-car-m3-starter-kit" || "$TARGET" == "r-car-h3-starter-kit" ]];  then
+  echo "Copying binary graphics and mmp drivers for $TARGET"
   cd "$BASEDIR/meta-renesas"
-  meta-rcar-gen3/docs/sample/copyscript/copy_evaproprietary_softwares.sh /var/go/sgx_bin_gen3/
+  meta-rcar-gen3/docs/sample/copyscript/copy_evaproprietary_softwares.sh $SGX_GEN_3_DRIVERS
   cd "$BASEDIR"
 fi
 


scripts/ci-build.sh
```diff
@@ -471,8 +471,9 @@ fi
 
 # Deal with special setup, copy binary drivers etc.
 if [[ "$TARGET" == "r-car-m3-starter-kit" || "$TARGET" == "r-car-h3-starter-kit" ]];  then
+  echo "Copying binary graphics and mmp drivers for $TARGET"
   cd "$BASEDIR/meta-renesas"
-  meta-rcar-gen3/docs/sample/copyscript/copy_evaproprietary_softwares.sh /var/go/sgx_bin_gen3/
+  meta-rcar-gen3/docs/sample/copyscript/copy_evaproprietary_softwares.sh $SGX_GEN_3_DRIVERS
```
gunnarx: Good catch

gunnarx:
Understood. It all looks good to me & I'll consider how to move this forward.  (The other layer updates are great but affect all targets due to the common master strategy, so we should retest).

slawr:
> (The other layer updates are great but affect all targets due to the common master strategy, so we should retest).

Understood. 

If you decide to prioritise the poky/meta-oe updates to later for some reason I would be happy for the YBSP v3.7.0 changes to be pulled ahead of them. They should still work and I am weary about someone doing a bunch of work on the GDP HMI issues and testing on R-Car using the older code.

gunnarx:
@slawr I Installed the new drivers on the build agents and it looks like the build was [successful](https://go.genivi.org/go/tab/build/detail/GDP-Yocto-r-car-m3-starter-kit-GITHUB-PR/288/build/2/init_and_bitbake).

**NOTE**: Ignore the "find" command at the top of the log.  It was set to the old sgx_bin_gen3 directory.  I removed it now, since it no longer makes sense.  Instead look further down, at what the copy script actuallly used.

gunnarx:
Let's try it, merged to rocko branch.

slawr:
>@slawr I Installed the new drivers on the build agents and it looks like the (M3) build was successful.
@gunnarx thanks for the results update and the merge. I did a visual check of the gfx/mmp copy script output of the build and its what I would expect.

So we have a solid improvement there going forward to be able to build different versions.

PR #196, opened 2018-08-16
meta-genivi-dev: remove linaro-gdp sublayer

mirzak:
This was something that was used in Yocto project version 2.3,
but was left dangling as the Yocto versions have been updated.

Safe to remove as there are no users.

[GDP-783] - Remove sublayer meta linaro-gdp


gunnarx:
Looks good to me.  Awaiting confirmation in JIRA regarding if also meta-linaro also should be removed at this point.  (If so, we include that too)

slawr:
R-Car CI build failure is due to setup problem. Failed to copy the gfx/mmp pkgs into the YBSP as it could not find any zip files in the source directory.

mirzak:
Yeah @gunnarx is making some updates in the pipelines to simplify the management of the firmware files.

That should be fixed soon.

PR #200, opened 2018-08-24
meta-genivi-dev: initial recipe for Apache MiNiFi

mirzak:
Proof of Concept state and have not tested enabling all the extensions,
but the default and core components build and deploy properly.

Tested running the minifi controller target with:

   /usr/bin/minifi.sh start

### DIFF
--- a/meta-genivi-dev/meta-genivi-dev/recipes-extended/minifi-cpp/minifi-cpp_0.5.0.bb
+++ b/meta-genivi-dev/meta-genivi-dev/recipes-extended/minifi-cpp/minifi-cpp_0.5.0.bb
@@ -1,0 +1,90 @@
+# NOTE!
+#
+# This recipe is only intended as a Proof of Concept. The reason being that
+# the minifi-cpp carries a lot of its dependency in the "thirdparty" directory.
+# Some of the thirdparty builds can be disabled and system libraries will be
+# used instead but most of the components in "thirdparty" are hardcoded to be
+# built and there is a lot of references to include directories in the various
+# cmake files to "thirdparty". This would need to involve the upstream project
+# to address which would allow to "system libraries" only.
+#
+# The proper clean-up would be to actually create a meta-minifi which provides
+# the recipes for all the dependencies that are not available in upstream
+# meta layers including recipe for minifi-cpp.
+#
+# Also note that not all compile time options are exposed in this recipe,
+# again this is only a PoC to compile and run MiNiFi on a target. TBD what is
+# actually to be done with this.
+
+
+SUMMARY = "Apache NiFi - MiNiFi C++ is a complementary data collection \
+approach that supplements the core tenets of NiFi in dataflow management, \
+focusing on the collection of data at the source of its creation. The C++ \
+implementation is an additional implementation to the one in Java with the \
+aim of an even smaller resource footprint."
+
+LICENSE = "MIT & GPLv2 & Apache-2.0 & EDL-1.0"
+LIC_FILES_CHKSUM = "\
+    file://LICENSE;md5=b58267cabd65636363f906f126095ff7 \
+    file://thirdparty/rapidjson-1.1.0/license.txt;md5=ba04aa8f65de1396a7e59d1d746c2125 \
+    file://thirdparty/uuid/COPYING;md5=58dcd8452651fc8b07d1f65ce07ca8af \
+    file://thirdparty/yaml-cpp-yaml-cpp-20171024/LICENSE;md5=6a8aaf0595c2efc1a9c2e0913e9c1a2c \
+    file://thirdparty/date/LICENSE.txt;md5=b5d973344b3c7bbf7535f0e6e002d017 \
+    file://thirdparty/spdlog-20170710/LICENSE;md5=6e5242b8f24d08c5e948675102937cc9 \
+    file://thirdparty/cxxopts/LICENSE;md5=8de00431559a76a1b43f6fd44f8f6689 \
+    file://thirdparty/civetweb-1.10/LICENSE.md;md5=da079d81be91ff1d1ca56dcd751f897f \
+    file://thirdparty/pcap++/LICENSE;md5=911690f51af322440237a253d695d19f \
+    file://thirdparty/pcap++/3rdParty/dirent-for-Visual-Studio/LICENSE;md5=b1177b216c5b9b15e89b9612a05ca461 \
+    file://thirdparty/pcap++/3rdParty/LightPcapNg/LightPcapNg/LICENSE.txt;md5=3b79e2c09be0f47e2543c923ea2a0bd7 \
+    file://thirdparty/libuvc-0.0.6/LICENSE.txt;md5=2f1963e0bb88c93463af750daf9ba0c2 \
+"
+
+SRC_URI = "\
+    https://github.com/apache/nifi-minifi-cpp/archive/rel/minifi-cpp-${PV}.tar.gz \
+    file://0001-MINIFICPP-562-Fixed-thirdparty-uuid-include-path.patch \
+    file://0002-thirdparty-date-remove-usage-of-SYSTEM.patch \
+    file://minifi.service \
+"
+
+SRC_URI[md5sum] = "2a63fd6692a55db2fedc99b986da32b4"
+SRC_URI[sha256sum] = "b7fb566706504c660ce43cd466ede58b649e94fb51f57dab8be02eabf6ea0f73"
+
+S = "${WORKDIR}/nifi-${BPN}-rel-${BPN}-${PV}"
+
+DEPENDS = "\
+    bison \
+    bzip2 \
+    curl \
+    flex \
+    icu \
+    openssl \
+    python3 \
+    zlib \
+"
+
+RDEPENDS_${PN} = "\
+    libcurl \
+"
+
+PACKAGECONFIG ?= "rocksdb"
+
+PACKAGECONFIG[mqtt] = "-DENABLE_MQTT=ON,-DENABLE_MQTT=OFF,paho-mqtt-c"
+PACKAGECONFIG[rocksdb] = "-DDISABLE_ROCKSDB=OFF,-DDISABLE_ROCKSDB=ON, rocksdb"
+PACKAGECONFIG[libarchive] = "-DDISABLE_LIBARCHIVE=OFF,-DDISABLE_LIBARCHIVE=ON,libarchive"
+PACKAGECONFIG[gps] = "-DENABLE_GPS=ON,-DENABLE_GPS=OFF,,libgps"
+
+EXTRA_OECMAKE = "-DSKIP_TESTS=ON -DBUILD_ROCKSDB=OFF -DUSE_SYSTEM_UUID=OFF"
+
+inherit cmake pkgconfig systemd
+
+do_install_append() {
+     rm ${D}/usr/NOTICE ${D}/usr/README.md ${D}/usr/LICENSE
+
+     install -D -m 0644 ${WORKDIR}/minifi.service \
+        ${D}${systemd_system_unitdir}/minifi.service
+}
+
+FILES_${PN} += "/usr/conf"
+FILES_${PN}-dev += "/usr/lib/cmake"
+
+SYSTEMD_SERVICE_${PN} = "minifi.service"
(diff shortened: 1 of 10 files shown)


gunnarx:
Looks good to me just to try out the component.  Normally we keep things clean but because of the state of this I appreciate that the caveats are written directly in the recipe comments, where they can easily be found if there are future conflicts and issues.  It facilitates creating a cleaner and more full-featured solution if the need arises.

gunnarx:
@mirzak The commit did not seem to add minifi-cpp component to actual build (not in packagegroup), so I amended and pushed it back, so that we get a build test.

mirzak:
@gunnarx, yeah good catch. I intentionally did not add it to any group as I was unsure if was to be included "by default", but I guess it could be.

gunnarx:
Then a style preference from me:

> meta-genivi-dev: initial recipe for Apache MiNiFi

IMHO we should drop the meta-genivi-dev: prefix in almost all commit descriptions, unless it really deals with some major reorganization of the layer itself.  Because that dir is the thing that primarily defines the GDP project itself, so it becomes mostly redundant information.  Almost every change or addition is done there, so prefixing it just adds noise to the interesting information.  Just think of it as modifying the GDP project rather than the meta-genivi-dev "layer".

instead of:
```
meta-genivi-dev:  <component>:  description
```

this is more useful, IMHO:
```
<component>: description
```

For changes in other layers, they first happen on those other repos so not the concern here.  Then when submodules are upgraded in GDP they are usually prefixed with the layer/submodule name, unless perhaps it's one specific component we're attempting to affect by the upgrade.  Changes in other dirs like gdp-src-build and scripts would normally be  prefixed by the core thing that is being changed.

```
<other layer>: update <...because...>
```

gunnarx:
> @gunnarx, yeah good catch. I intentionally did not add it to any group as I was unsure if was to be included "by default", but I guess it could be.

Sure but we need a build test.  I'd say a refactor of the groups and variants is fairly imminent so that at least one variant uses it.

meta-genivi-dev/meta-genivi-dev/recipes-extended/minifi-cpp/minifi-cpp_0.5.0.bb
```diff
@@ -0,0 +1,81 @@
...
+PACKAGECONFIG ?= "rocksdb"
+
+PACKAGECONFIG[mqtt] = "-DENABLE_MQTT=ON,-DENABLE_MQTT=OFF,paho-mqtt-c"
+PACKAGECONFIG[rocksdb] = "-DDISABLE_ROCKSDB=OFF,-DDISABLE_ROCKSDB=ON, rocksdb"
+PACKAGECONFIG[libarchive] = "-DDISABLE_LIBARCHIVE=OFF,-DDISABLE_LIBARCHIVE=ON,libarchive"
+PACKAGECONFIG[gps] = "-DENABLE_GPS=ON,-DENABLE_GPS=OFF,"
+
+RDEPENDS_${PN} += "${@bb.utils.contains('PACKAGECONFIG', 'gps', 'libgps', '', d)}"
+
+EXTRA_OECMAKE = "-DSKIP_TESTS=ON -DBUILD_ROCKSDB=OFF -DUSE_SYSTEM_UUID=OFF"
```
chbae: why don't you define this as PACKAGECONFIG?

meta-genivi-dev/meta-genivi-dev/recipes-extended/minifi-cpp/minifi-cpp_0.5.0.bb
```diff
@@ -0,0 +1,81 @@
...
+
+inherit cmake pkgconfig
+
+do_install_append() {
+     rm ${D}/usr/NOTICE ${D}/usr/README.md ${D}/usr/LICENSE
```
chbae: please add -rf option not to get the error in second build time.

meta-genivi-dev/meta-genivi-dev/recipes-extended/minifi-cpp/minifi-cpp_0.5.0.bb
```diff
@@ -0,0 +1,81 @@
...
+}
+
+FILES_${PN} += "/usr/conf"
+FILES_${PN}-dev += "/usr/lib/cmake"
```
chbae: not required. it already defined in cmake.bbclass

meta-genivi-dev/meta-genivi-dev/recipes-extended/minifi-cpp/minifi-cpp_0.5.0.bb
```diff
@@ -0,0 +1,81 @@
...
+"
+
+PACKAGECONFIG ?= "rocksdb"
+
+PACKAGECONFIG[mqtt] = "-DENABLE_MQTT=ON,-DENABLE_MQTT=OFF,paho-mqtt-c"
+PACKAGECONFIG[rocksdb] = "-DDISABLE_ROCKSDB=OFF,-DDISABLE_ROCKSDB=ON, rocksdb"
+PACKAGECONFIG[libarchive] = "-DDISABLE_LIBARCHIVE=OFF,-DDISABLE_LIBARCHIVE=ON,libarchive"
+PACKAGECONFIG[gps] = "-DENABLE_GPS=ON,-DENABLE_GPS=OFF,"
+
+RDEPENDS_${PN} += "${@bb.utils.contains('PACKAGECONFIG', 'gps', 'libgps', '', d)}"
```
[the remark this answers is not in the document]
mirzak: Yeah, true.

meta-genivi-dev/meta-genivi-dev/recipes-extended/minifi-cpp/minifi-cpp_0.5.0.bb
```diff
@@ -0,0 +1,81 @@
...
+    curl \
+    flex \
+    icu \
+    openssl \
+    python3 \
+    zlib \
+"
+
+RDEPENDS_${PN} = "\
+    libcurl \
```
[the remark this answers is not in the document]
gunnarx: Hey @chbae - thanks for reviewing! :)  I think this article was quite good. https://flameeyes.blog/2008/05/15/when-you-should-use-rdepend-and-when-you-should-use-depend/

meta-genivi-dev/meta-genivi-dev/recipes-extended/minifi-cpp/minifi-cpp_0.5.0.bb
mirzak: This are not optional or configurable as the package will not build correctly with these settings. Overall the upstream cmake files is a mess and would require a lot of work to actually write a "proper" recipe. Something that we decided not to prioritize right now, and might come back to it when we have defined use-case for MiNiFi.

meta-genivi-dev/meta-genivi-dev/recipes-extended/minifi-cpp/minifi-cpp_0.5.0.bb
mirzak: Nope, this is by design and should fail if one of the files does not exist. This will indicate that something has changed in the upstream package.

There is a dependency chain in place here, meaning that do_install_append will always run after do_install meaning that these files will always be present on the current revision. There will not be any problems with "second build" because if the package is re-built that means that do_install will run, or if the sstate-cache is re-used then it wont run any of the steps.

meta-genivi-dev/meta-genivi-dev/recipes-extended/minifi-cpp/minifi-cpp_0.5.0.bb
mirzak: I can not see anything that adds this in cmake.bbclass and please provide a reference in case I missed but.

But this is required,  otherwise there will be "installed vs shipped" errors.

meta-genivi-dev/meta-genivi-dev/recipes-extended/minifi-cpp/minifi-cpp_0.5.0.bb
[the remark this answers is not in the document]
mirzak: Because the runtime dependency is on `libcurl` and not on `curl`, those are not the same thing. 

I will drop `curl` from DEPENDS as it was not needed anyway.

mirzak:
And of course, @chbae thank you allocating time for review. Highly appreciated.

meta-genivi-dev/meta-genivi-dev/recipes-extended/minifi-cpp/minifi-cpp_0.5.0.bb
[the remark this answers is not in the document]
mirzak: Actually `curl`is needed in DEPENDS, so left it as-is

mirzak:
Updated MiNiFI recipe based on feedback from @chbae and @gunnarx.

Also added some addition comments at the top of the recipe to clarify current state

mirzak:
Updated with systemd service that will auto-start MiNiFi.

meta-genivi-dev/meta-genivi-dev/recipes-extended/minifi-cpp/minifi-cpp_0.5.0.bb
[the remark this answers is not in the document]
chbae: libcurl is for library installation or other binary?
It's very rare to add library in RDEPENDS because it's for runtime dependency.

meta-genivi-dev/meta-genivi-dev/recipes-extended/minifi-cpp/minifi-cpp_0.5.0.bb
chbae: make sense. :)

meta-genivi-dev/meta-genivi-dev/recipes-extended/minifi-cpp/minifi-cpp_0.5.0.bb
chbae: sorry. that's my mistake. you may be correct.

mirzak:
The test failure is because the disk is full when running qemu, it overflowed when MiNiFi was added. And it fails because on SSH test because it can not create identity files. Great tests we have :D

Though can not figure out where it picks up the "disk size" as it does not seem to have pick this up from the ext4 file size.

mirzak:
Needs https://github.com/GENIVI/genivi-dev-platform/pull/203 and https://github.com/GENIVI/meta-ivi/pull/84 to be able to pass the acceptance tests.

meta-genivi-dev/meta-genivi-dev/recipes-extended/minifi-cpp/minifi-cpp_0.5.0.bb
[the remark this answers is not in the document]
gunnarx: But isn't it better to explicitly write what is needed during runtime?  And use RDEPENDS for that, not to rely on what DEPENDS brings in by chance?

> It's very rare to add library in RDEPENDS because it's for runtime dependency.

I did not understand this, because during runtime it is necessary to load libraries, right?

The article I linked to says:

_"RDEPEND instead is the set of packages that has to be available during execution of the program. This means it has to have the libraries linked against, if shared "_

Anyway, I'll assume this is OK now and merge.  If you have a proposal for change, please send it as a new patch/PR and we can discuss more.

gunnarx:
> Needs #203 and GENIVI/meta-ivi#84 to be able to pass the acceptance tests.

OK #203 is merged and #84 was merged earlier so I bumped meta-ivi to the tip of 14.x-rocko branch.
There are quite a lot of changes there, so I'll push an update to rerun the build checks once more.

mirzak:
It is expected for the LAYER_CONF_VERSION needing update because of this change, https://github.com/GENIVI/meta-ivi/commit/ed8ada9064caf866c55a473aeb69e3682a4877a7.

The LAYER_CONF_VERSION entry that was in poky-ivi-systemd.conf was a bit outdated, and as it now inherited from poky.conf it is set to version 7.

gunnarx:
GitHub says it's conflicting but git rebased this on master without issue. Hmm... 
Guess I have to push an update.  That means the build checks won't be visible so just to document
https://github.com/GENIVI/genivi-dev-platform/pull/200/commits/36f80b1737b437903ce3410823d7a549a5865de4 (minifi-cpp: No security/hardening cflags)  build tested minnowboard, qemu + test, r-pi3 ...

PR #203, opened 2018-09-07
Qemu fixes

mirzak:
First patch fixes an issue where disk could end up full when booting QEMU.

Second is a "stability improvement"


gunnarx:
Good stuff.

PR #204, opened 2018-09-12
bblayers.inc : Update LCONF_VERSION to 7

gunnarx:
poky is now on LAYER_CONF_VERSION=7 and for some reason the built-in automatic upgrade of LCONF_VERSION has started to fail execution inside sanity.bbclass.   Some python/parsing issue. It might have a fix already.  Anyway I can't figure it out and thought we can just as well update our conf version.

It is time to update poky (rocko maintenance branch) also anyway but that can come after this.

### DIFF
--- a/gdp-src-build/conf/templates/bblayers.inc
+++ b/gdp-src-build/conf/templates/bblayers.inc
@@ -1,6 +1,6 @@
 # LAYER_CONF_VERSION is increased each time build/conf/bblayers.conf
 # changes incompatibly
-LCONF_VERSION = "6"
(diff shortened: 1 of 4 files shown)


gunnarx:
This PR was used mostly as a stepwise check of some of the patches.  All of these fixes are required also in #200 since it too updates the baseline layer.  #200 is now rebased to include it and therefore this PR is basically obsolete.  When #200 is merged, double check there is nothing missing, and this can then be closed.

gunnarx:
Obsolete since #200 includes it and #200 was merged.

PR #205, opened 2018-09-13
Remove iotivity

gunnarx:
At the moment the meta-layers for iotivity are quite out of date and we don't have the bandwidth to update them.  There are also some immediate compilation issues right now that is holding back progress on
other fronts so we have to consider dropping it from master.

I've had some contact with previous and new maintainers, but it is still lagging behind.  We can always reverse this decision and it is possible it will remain on a certain branch/variant of GDP or that we
get some sort of update later on.

Let's see what the effect of this is.  We might decide to merge this, at least for the moment.
I am using the JIRA ticket to document more about this [GDP-809](https://at.projects.genivi.org/jira/browse/GDP-809)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -13,13 +13,6 @@
 [submodule "poky"]
 	path = poky
 	url = git://git.yoctoproject.org/poky
-[submodule "meta-oic"]
-	path = meta-oic
-	url = http://git.yoctoproject.org/git/meta-oic.git
-	branch = 1.0.1
-[submodule "meta-iot-web"]
-	path = meta-iot-web
-	url = https://github.com/ostroproject/meta-iot-web
 [submodule "meta-rvi"]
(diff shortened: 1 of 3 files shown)


gunnarx:
Documenting that build checks (3 targets) and self test successful for 24d978d017475212a16ad47e93a1c5dc796fa635. 

Despite improvements in git, there can be some issues when fetching a remote change that removes a submodule - especially if it's an old version working directory.   git deinit <submodule> should usually help, as far as I understand.

For the go-agents, they incrementally update all material (git repos) rather than doing full new clones, which is a good feature, but submodule removal like this has also been an issue there on some existing pipelines.  In those cases we need to remove the material so a fresh clone is created.

gunnarx:
Check if this is obsolete once #200 is merged.

PR #206, opened 2018-09-13
ci-build: Extract the interesting log files from Yocto build

gunnarx:
For CIAT usage. Yocto/bitbake often refers to a detailed log for the particular compilation step for a particular compopnent (e.g. log.do_compile.29015).  We want to capture in particular the logs that were referenced as failures, and make them available for display in the CIAT infrastructure. 

At first I had a separate script but for simplicity I added it to the ci-build script instead.

### DIFF
--- a/scripts/ci-build.sh
+++ b/scripts/ci-build.sh
@@ -219,6 +219,60 @@ cleanup() {
   fi
 }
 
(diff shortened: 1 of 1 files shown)


gunnarx:
This will not be updated.

PR #207, opened 2018-09-13
Minor cleanups in renesas local.conf templates

mirzak:


gunnarx:
Looks good to remove some duplication, but I'll defer to slawr for further review.

slawr:
> But this all gen2 boards right? Which are not effect by my patch at all because the change was moved to a file that is only common to "gen3" boards (renesas-rcar-gen3.local.inc).

Nope. Gen 3 SoCs = e.g H3, M3, D3, V3H. H3/M3 Starter Kit and H3/M3/M3N Salvator-X(S) are all Gen 3 boards.

mirzak:
Ok, will drop the second patch.

gunnarx:
@mirzak  Should 41613a3 be dropped completely then?  Or did you already adjust this...  
(I can drop it when merging, as long as I know)

mirzak:
Yeah, 41613a3 can completely be dropped. You can do that when merging. But still need ACK from @slawr on the other commit.

gunnarx:
@slawr can I merge the 5ac77d6 commit (only)?

slawr:
Hi, had to concentrate on some topics. Plan to catch up on some things including this in the next day or two. I need to check my notes to see why I added it to the per machine conf.

Has it been tested to check the correct value is set without it?

gunnarx:
> Has it been tested to check the correct value is set without it?

Not sure.  How can we test it?

slawr:
> Not sure. How can we test it?

It's a variable so you can just bitbake environment dump to check that var is set correctly without that line - that's Mirza's hypothesis that it is a duplication. So check the hypothesis.

If you wanted to go onto "belt and braces" you could sanity test an M3 SK build.

mirzak:
I did do a sanity check of the variables and a build of the M3 SK. I am not throwing out pull-requests without at least some sanity testing :).

We should have also gotten a build from our CI, but can only see minnowboard though.

mirzak:
SOC_FAMILY is already setup in meta-renesas,

https://github.com/renesas-rcar/meta-renesas/blob/rocko/meta-rcar-gen3/conf/machine/m3ulcb.conf#L5
https://github.com/renesas-rcar/meta-renesas/blob/rocko/meta-rcar-gen3/conf/machine/h3ulcb.conf#L5

so yeah setting MACHINE = "m3ulcb", will result in SOC_FAMILY being set to what is defined in m3ulcb.conf

mirzak:
Updated PR where I have dropped the second commit as agreed.

slawr:
For the life of me I can't find the review "approve" button.

gunnarx:
I'll take that as approval :)

PR #212, opened 2018-10-03
R-car gen2 unsupported, and gen3 improvements 

gunnarx:
@slawr for a quick review

At first I was uncertain which $MACHINE to set for salvator-x and if it depended on the m3/h3 choice (like it does for starter-kit).  But from what I can tell, it should work to set MACHINE that only in the conf-files as I think we discussed at some point.  So, after the last commit it is not a question I need to answer. 
However, still need to add proper template files for salvator-x, and if we were going to do some local modification for eva drivers, we add that to the script.


scripts/ci-build.sh
```diff
@@ -261,16 +261,22 @@ define_with_default KEEP_TMP false
 
 stop_if_failure
 
-git_gdp="https://github.com/GENIVI/genivi-dev-platform"
```
slawr: Nit pick: This section seems unrelated to adding salvator-x, i.e. should it be stand alone "clean-up".

gunnarx:
* You are right -  I was first mostly thinking of what is "supported" from the point of view of our defined build (which is defined by the scripts + instructions).  But I think now it makes little sense to keep the old Gen 2 files around so I updated that commit to remove all references I could find.  There were some remaining in READMEs also.  
* Fixed title to use full name of board
* Agreed, cleanup does not need to be mixed into the commits for Renesas.  Added now as part of the other cleanup commit which is close enough.  (The PR itself is mixed, yes, but I'm not quite as picky about that)

gunnarx:
#219 has been incorporated here

gunnarx:
Amended commit 332cced and pushed update.

gunnarx:
* The non-Renesas builds were stopped and therefore show failure.  

* Restarted starter-kit build to ensure nothing has broken there (there should be little risk since it uses its own conf/template files).  Result is OK

* Salvator builds OK with this commit (0ae699f) : [here](https://go.genivi.org/go/pipelines/GDP.r-car-m3-salvator-x/4/build/1) and [here](https://go.genivi.org/go/pipelines/GDP.r-car-h3-salvator-x/7/build/1).  Have no reason to think it's not running correctly as well but feel free to test & report results.

PR #213, opened 2018-10-04
Image profiles

mirzak:
Two new image profiles have been introduced:

- genivi-dev-platform-core
    - headless minimal image
- genivi-dev-platform-data
    - big data / telematics

`genivi-dev-platform` has stayed as-is with one difference, it now inherits some functionality from `genivi-dev-platform-core` image. The intention is not to change the functionality of it in anyway at this stage.


gunnarx:
Your code is so good so we need to nitpick ;)

Commit messages: 
1. According to [GDP Rules for Commit Messages](https://at.projects.genivi.org/wiki/x/ZwHe) your lines are too long.  (Pretty much the same rules as many other projects, e.g. [GNOME commit rules](https://wiki.gnome.org/Git/CommitMessages) but that specifies 75 characters and for some reason we specify 74...)   But you seem to wrap at 80 currently, so you should try to adjust that.

2. The text "minimum for in image" (in more than one commit) sounds strange.

gunnarx:
Looking good now.  I had not saved my review for you to see the image description comment in time.
But I could tweak the image descriptions during the merge to my own liking ;)

gunnarx:
OK, so the build checks show failure because they were stopped.  The PR pipeline don't really test that the two new variants work.  I set up temporary separtae pipelines to test core and data.

PR #215, opened 2018-10-05
adopt "UsrMerge" philosophy (partly)

mirzak:
Read more about [UsrMerge](https://www.freedesktop.org/wiki/Software/systemd/TheCaseForTheUsrMerge/).

This is actually required for libostree and these are changes that we have in our [libostree](https://github.com/GENIVI/genivi-dev-platform/tree/libostree) branch that makes sense to merge even without the actual libostree support.

It is a set of commits that changes install paths to `/usr` instead of "all over the place".

While doing above I did big revamp of HMI application recipe structure as there was a lot of duplication among the recipes.


gunnarx:
Agreed - implementing "UsrMerge" makes sense for many reasons, and a lot of these recipes were all over the place, just needing someone to do some cleanup.

gunnarx:
qemu was failing to start - some problem upgrade - maybe a mismatch host/container.  In any case an upgrade of qemu (inside go container) solved it.   

I re-triggered test stage.  The test suite seems to run OK.

One test still fails on RVI service not starting (due to changed paths).  Should be a simple fix but is still an open issue we did not bring to conclusion.  Ref:  https://github.com/GENIVI/rvi_core/pull/116

gunnarx:
That did the trick.  Automated tests now run OK.  salvator-x build fail - likely it just needs rebasing on latest master fixes. Rebasing & merging.

gunnarx:
That did the trick.  Automated tests now run OK.  salvator-x build fail - likely it just needs rebasing on latest master fixes. Rebasing & merging.

PR #218, opened 2018-10-05
Test Remove meta-gnome layer

gunnarx:
Looks like chromium depended on gnome keyring but after the switch to
meta-lgsvl-browser layer, I can't find this, or any dependency any longer.  
Let's see if it builds without meta-gnome.


gunnarx:
Apparently there's a remaining dependency chain sdl -> bluez-tools -> obexd (which exists in meta-gnome)

PR #219, opened 2018-10-05
init: add Renesas R-Car Gen 3 Salvator-X(S) board support

slawr:
Add missing local.conf and bblayer.conf support for Salvator-X to PR #212. Needs to be merged.
Untested. Would set WIP label but I don't seem to have the rights to set them for some reason.


gunnarx:
Was merged as part of #212

PR #222, opened 2018-10-26
cannelloni: fix license checksum

slawr:
The license file md5 checksum is specified in uppercase which can lead
to a build error when compared to a lowercase sum generated at runtime.
For example:
ERROR: cannelloni-1.0-r0 do_populate_lic: QA Issue: cannelloni: The LIC_FILES_CHKSUM does not match for file://gpl-2.0.txt;md5=B234EE4D69F5FCE4486A80FDAF4A4263
cannelloni: The new md5 checksum is b234ee4d69f5fce4486a80fdaf4a4263

To avoid such issues specify the checksum in lowercase.

The strange thing in my mind is why this has not been an error before as the checksum check in insane.bbclass appears to be case sensitive. I see the error in a build for H3 Starter Kit, but M3 Starter Kit is ok. I've posted a question about that to the YP mailing list.

### DIFF
--- a/meta-genivi-dev/meta-genivi-dev/recipes-extended/cannelloni/cannelloni.bb
+++ b/meta-genivi-dev/meta-genivi-dev/recipes-extended/cannelloni/cannelloni.bb
@@ -1,7 +1,7 @@
 SUMMARY = "SocketCAN over Ethernet tunnel using UDP to transfer CAN frames between two machines"
 SECTION = "console/network"
 LICENSE = "GPLv2"
-LIC_FILES_CHKSUM = "file://gpl-2.0.txt;md5=B234EE4D69F5FCE4486A80FDAF4A4263"
+LIC_FILES_CHKSUM = "file://gpl-2.0.txt;md5=b234ee4d69f5fce4486a80fdaf4a4263"
 
 PR = "r0"
 


slawr:
@gunnarx M3 SK check failed due to missing gfx pkgs. Job assigned to the wrong agent maybe? H3 Salvator-X check got past that stage.

gunnarx:
> Job assigned to the wrong agent maybe

Nah, it was my mistake.  I recreated one agent container from scratch yesterday and forgot to run the Fetch_rcar_binaries job that redistributes the binaries to all agents.  All sorted now and the pipeline is [restarted](https://go.genivi.org/go/tab/build/detail/GDP-Yocto-r-car-m3-starter-kit-GITHUB-PR/421/build/2/init_and_bitbake)

gunnarx:
By the way... this already fixed on GDP sumo branch (0f50561752cf40e60ba32e251377c14887a77d68)  Sorry that we're not yet merged...

I think some QA checks have been there for a while but are now treated as errors in recent yocto/bitbake upates.  

It's possible you ran into this while upgrading some layers?   
(It's also possible that this problem exists also on master and we did not notice it because of caching.  In any case sumo should be merged soon)

slawr:
> By the way... this already fixed on GDP sumo branch (0f50561) Sorry that we're not yet merged...
No unfortunately didn't see that :( Was that prompted by my report some time ago or an issue you say yourself? Interested mostly in if its the latter as curious as to why it was not happening all the time.

> I think some QA checks have been there for a while but are now treated as errors in recent yocto/bitbake upates.

I didn't get as far as matching the poky shas for both the working and non-working builds and then tracing down any differences. Guess they might have turned on something but both would have been using very similar environments.

> 
> It's possible you ran into this while upgrading some layers?
> (It's also possible that this problem exists also on master and we did not notice it because of caching. In any case sumo should be merged soon)

When I originally reported some weeks back it was a fresh clone but yes this time around almost certainly did a pull of upstream to test the latest.

Didn't get any replies to my email to the YP ML. Guess it is a bit of a prosiac query. Thought I might learn something about python at least and maybe there was something deeper to learn for YP potentially. But bigger fish to fry. So moving on.. :)

gunnarx:
> Was that prompted by my report some time ago or an issue you saw yourself?

Don't know. Sorry, the trigger is often not something I would commit to memory.  I just fix and move on...  

If you prefer we can merge this immediately and I'll drop the other one when rebasing the sumo branch.

> Interested mostly in if its the latter as curious as to why it was not happening all the time

I hope a reasonable explanation for that was given in my previous comment?  In any case for me it was enough that it happened one time... so it was fixed.  That's as much as I am going to worry about it at least ;-)

slawr:
Ok agree to move on.

Don't mind whether you take this or the one on sumo. We just need it on master :)

gunnarx:
OK - since you say we need it on master, the quickest now is to just merge this one
I'll handle the conflict during sumo merge.

PR #224, opened 2018-11-01
reflect latest tagged release on master

mirzak:
14.1.x-sumo is ahead of master currently but it is also our latest tagged release, merge it back so that any further work on master is based on the latest release.


mirzak:
Aborting, we will bring this in using a merge commit as there are changes on master not present `14.1.x-sumo`

PR #226, opened 2018-11-01
Salvator-x fixes to rocko maintenance branch

gunnarx:
Sending this as PR to check the content.  Commits needed to be cherry picked separately since master has moved ahead to sumo.  These commits were proposed to master before that, so the intention should be that they apply also for rocko.  Just double-checking with @slawr and getting a build test on this first.


slawr:
@gunnarx can we merge this now. I'm having to move from sanity testing Salvator-X on Master branch due to the switch to Sumo there and the rocko branch is out of date :(

PR #227, opened 2018-11-02
gdp-new-hmi: fix SRCREV

mirzak:
This commit:

    https://github.com/GENIVI/genivi-dev-platform/commit/fa227829f1696775a02396eafb73db05748b4f1f

Accidently downgraded the GDP HMI to a revision that is
broken. Fix by updating to the latest known revision.


gunnarx:
That hash has been tested before and it seems to build, so I'll just go ahead and merge.  This is a hotfix as far as I can see it.

gunnarx:
My bad not catching it in the review also...

Anyway, merged and backported to 14.0.x-rocko and 14.1.x-sumo maintenance branches which also had the bug for a little while now, but fortunately not the tagged versions v14.0 and v14.1

